PR #151, opened 2021-07-15
fix: org name validation

Kirpal:
## Description
When trying to use the new `--org-name` support I found that underscores were not being supported. After looking a little closer, the regex for the org name was incorrect by not enforcing that segments must start with a letter and by enforcing a min/max of three segments.

I updated the org name regex, invalid org error message, and added more tests for edge cases.

#### Sources:
- [Android rules](https://developer.android.com/studio/build/application-id)
- [iOS rules](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier#discussion)
- [Flutter android validation](https://github.com/flutter/flutter/blob/76d5e62f42dffe3aba42bba9364c4469af1ff332/packages/flutter_tools/lib/src/commands/create_base.dart#L481)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/commands/create.dart
+++ b/lib/src/commands/create.dart
@@ -137,12 +137,16 @@ class CreateCommand extends Command<int> {
   }
 
   /// Gets the organization name.
-  List<String> get _orgName {
-    if (_argResults['org-name'] == null) return _defaultOrgName.split('.');
-
-    final orgName = _argResults['org-name'] as String;
+  List<Map<String, String>> get _orgName {
+    final orgName = _argResults['org-name'] as String? ?? _defaultOrgName;
     _validateOrgName(orgName);
-    return orgName.split('.');
+    final segments = orgName.replaceAll(RegExp(r'-|_'), ' ').split('.');
+    return segments.map((segment) {
+      return {
+        'value': segment,
+        'separator': segment == segments.last ? '' : '.'
+      };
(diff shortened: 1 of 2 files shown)


felangel (changes requested):
Before this can be merged we will need to update the template to support dynamic length for org name parts.

lib/src/commands/create.dart
```diff
@@ -151,8 +150,10 @@ class CreateCommand extends Command<int> {
     if (!isValidOrgName) {
       throw UsageException(
         '"$name" is not a valid org name.\n\n'
-        'A valid org name has 3 parts separated by "."'
-        'and only includes alphanumeric characters and underscores'
+        'A valid org name has:\n'
+        '  - At least 2 parts separated by "."\n'
+        '  - Each part must start with a letter\n'
+        '  - Only includes alphanumeric characters and underscores'
```
felangel, line 156: According to the iOS spec:

> only alphanumeric characters (A-Z, a-z, and 0-9), hyphens (-), and periods (.)

I don't think we should allow underscores.

Kirpal: That's a great idea, I'll update this

felangel (approved):

PR #188, opened 2021-09-22
refactor: modularize cli utilities

felangel:
## Description

- refactor: modularize cli utilities
  - preparation for adding a Dart CLI which supports `dart fix`

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/cli/cli.dart
+++ b/lib/src/cli/cli.dart
@@ -12,8 +12,12 @@ class _Cmd {
     bool throwOnError = true,
     String? workingDirectory,
   }) async {
-    final result = await Process.run(cmd, args,
-        workingDirectory: workingDirectory, runInShell: true);
+    final result = await Process.run(
+      cmd,
+      args,
+      workingDirectory: workingDirectory,
+      runInShell: true,
+    );
 
     if (throwOnError) {
       _throwIfProcessFailed(result, cmd, args);
--- a/lib/src/cli/cli.dart
+++ b/lib/src/cli/cli.dart
@@ -1,0 +1,43 @@
+import 'package:path/path.dart' as p;
+import 'package:universal_io/io.dart';
+
+part 'flutter_cli.dart';
+
+/// Abstraction for running commands via command-line.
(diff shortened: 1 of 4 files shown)


marcossevilla:
@felangel what if we use barrel files for a "CLI feature" instead of using `part of`, as they might be more CLIs in the future and might be easier to maintain as we would have imports specific to one CLI on each file?

Either approach looks good to me but just had that thought. 🤔

felangel:
> @felangel what if we use barrel files for a "CLI feature" instead of using `part of`, as they might be more CLIs in the future and might be easier to maintain as we would have imports specific to one CLI on each file?
> 
> Either approach looks good to me but just had that thought. 🤔

The reason I went with this approach is to keep `_Cmd` private. If we went with a barrel we'd need to publicly expose the `Cmd` class.

lib/src/cli/cli.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:path/path.dart' as p;
+import 'package:universal_io/io.dart';
+
+part 'flutter_cli.dart';
+
+/// Abstraction for running commands via command-line.
+class _Cmd {
+  /// Runs the specified [cmd] with the provided [args].
+  static Future<ProcessResult> run(
```
orestesgaolin, line 9: Are you fully convinced to run commands via static methods? Maybe having a CommandClient or something similar would be more inline with our general approach to modularization?

felangel: Yeah that's a larger refactor we can prioritize. My primary goal for this PR was to make it so we could easily introduce a `dart_cli.dart` to support `dart fix`. The existing `_Cmd` API is not affected by these changes but I think we can create a separate PR to address that. wdyt?

orestesgaolin: sounds good, just wanted to chime in but I'm ok with the current approach

orestesgaolin (approved):

PR #195, opened 2021-09-28
feat(flutter_plugin): conditional platform support

marcossevilla:
## Description

- Update flutter_plugin mason bundle to the latest version with conditional folder support.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
shouldn't we add flags for each platform so that developers can specify which platforms they want? Currently all platforms will just default to false.

marcossevilla:
> shouldn't we add flags for each platform so that developers can specify which platforms they want? Currently all platforms will just default to false.

should be good now, let me know what you think. 👍

lib/src/commands/create.dart
```diff
@@ -106,12 +136,24 @@ class CreateCommand extends Command<int> {
     final template = _template;
     final generateDone = _logger.progress('Bootstrapping');
     final generator = await _generator(template.bundle);
+    final android = _argResults['android'] as String? ?? 'true';
+    final ios = _argResults['ios'] as String? ?? 'true';
+    final web = _argResults['web'] as String? ?? 'true';
+    final linux = _argResults['linux'] as String? ?? 'true';
+    final macos = _argResults['macos'] as String? ?? 'true';
+    final windows = _argResults['windows'] as String? ?? 'true';
```
felangel, line 144:
```suggestion
    final android = _argResults['android'] as bool? ?? true;
    final ios = _argResults['ios'] as bool? ?? true;
    final web = _argResults['web'] as bool? ?? true;
    final linux = _argResults['linux'] as bool? ?? true;
    final macos = _argResults['macos'] as bool? ?? true;
    final windows = _argResults['windows'] as bool? ?? true;
```

lib/src/commands/create.dart
```diff
@@ -106,12 +136,24 @@ class CreateCommand extends Command<int> {
...
         'project_name': projectName,
         'description': description,
-        'org_name': orgName
+        'org_name': orgName,
+        'android': android.parseBool(),
+        'ios': ios.parseBool(),
+        'web': web.parseBool(),
+        'linux': linux.parseBool(),
+        'macos': macos.parseBool(),
+        'windows': windows.parseBool(),
```
felangel, line 156:
```suggestion
        'android': android,
        'ios': ios,
        'web': web,
        'linux': linux,
        'macos': macos,
        'windows': windows,
```

lib/src/commands/create.dart
```diff
@@ -220,3 +262,9 @@ class CreateCommand extends Command<int> {
     }
   }
 }
+
+/// Extension on String to parse booleans.
+extension StringToBoolX on String {
+  /// Returns if `String` is equal to `true`.
+  bool parseBool() => toLowerCase() == 'true';
+}
```
felangel, line 270:
```suggestion
```

lib/src/commands/create.dart
marcossevilla: I'm rolling back to the previous approach because ci fails to cast [here](https://github.com/VeryGoodOpenSource/very_good_cli/pull/195/checks?check_run_id=3760664902).

felangel (approved):

PR #296, opened 2022-02-23
refactor: upgrade to mason 0.1.0-dev.8

felangel:
## Description

- refactor: upgrade to mason 0.1.0-dev.8

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/test/src/commands/create_test.dart
+++ b/test/src/commands/create_test.dart
@@ -76,6 +76,11 @@ void main() {
     late PubUpdater pubUpdater;
     late VeryGoodCommandRunner commandRunner;
 
+    final generatedFiles = List.filled(
+      62,
+      const GeneratedFile.created(path: ''),
+    );
+
     void Function() overridePrint(void Function() fn) {
       return () {
         final spec = ZoneSpecification(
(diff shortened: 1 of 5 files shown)


test/src/commands/create_test.dart
```diff
@@ -484,7 +484,7 @@ void main() {
             ),
           ).thenAnswer((_) async {
             File(p.join('.tmp', 'pubspec.yaml')).writeAsStringSync(pubspec);
-            return 62;
+            return List.filled(62, const GeneratedFile.created(path: ''));
```
erickzanardo, line 487: Nit: this is being repeated on the test, maybe would be worth adding it to a method for the sake of DRY

erickzanardo (approved):
One minor nit, but LGTM already 🔥

test/src/commands/create_test.dart
felangel: Yeah I'm fine either way the reason I left it as is is because I don't mind repeating code in tests in order to keep the tests self-contained/easy to understand. Happy to adjust if you prefer we create a helper for this. Let me know what you had in mind 👍

erickzanardo: I was thinking on a generateFileList which just receive the length, so some details of how the file is mocked is not repeated.

Either way is just a nit, I know DRY in tests is a complicated matter hahahha

felangel: I updated it a bit, lmk what you think

erickzanardo (approved):

test/src/commands/create_test.dart
erickzanardo: LGTM! 🚀

PR #308, opened 2022-03-09
feat: use very_good_test_runner to improve test output

felangel:
## Description

- feat: use very_good_test_runner to improve test output

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


scarletteliza:
my heart is going pitter patter

lib/src/cli/flutter_cli.dart
```diff
@@ -65,55 +62,182 @@ class Flutter {
   static Future<void> test({
     String cwd = '.',
     bool recursive = false,
-    void Function([String?]) Function(String message)? progress,
-  }) async {
-    await _runCommand(
-      cmd: (cwd) async {
-        final installDone = progress?.call(
-          'Running "flutter test" in $cwd',
+    void Function(String)? stdout,
```
erickzanardo, line 65: Nit: Can we create a typedef for this?

lib/src/cli/flutter_cli.dart
```diff
@@ -65,55 +62,182 @@ class Flutter {
...
       },
       cwd: cwd,
       recursive: recursive,
     );
   }
+}
+
+/// Run a command on directories with a `pubspec.yaml`.
+Future<void> _runCommand<T>({
+  required Future<T> Function(String cwd) cmd,
```
erickzanardo, line 86: Nit: maybe have a type def for this?

lib/src/cli/flutter_cli.dart
```diff
@@ -65,55 +62,182 @@ class Flutter {
...
-      return;
-    }
+  if (processes.isEmpty) throw PubspecNotFound();
 
-    final processes = _Cmd.runWhere(
-      run: (entity) => cmd(entity.parent.path),
-      where: _isPubspec,
-      cwd: cwd,
-    );
+  for (final process in processes) {
```
erickzanardo, line 106: Do we need they to be in sequence? If not we could gain time by running in parallel?

lib/src/cli/flutter_cli.dart
```diff
@@ -65,55 +62,182 @@ class Flutter {
...
+    await process;
+  }
+}
 
-    if (processes.isEmpty) throw PubspecNotFound();
+Future<void> _flutterTest({
+  String cwd = '.',
+  required void Function(String) stdout,
```
erickzanardo, line 113: Nit: reuse the typedef here

lib/src/cli/flutter_cli.dart
```diff
@@ -65,55 +62,182 @@ class Flutter {
...
+extension on Duration {
+  String formatted() {
+    String twoDigits(int n) => n.toString().padLeft(2, '0');
+    final twoDigitMinutes = twoDigits(inMinutes.remainder(60));
+    final twoDigitSeconds = twoDigits(inSeconds.remainder(60));
+    return darkGray.wrap('$twoDigitMinutes:$twoDigitSeconds')!;
+  }
+}
+
+extension on int {
```
erickzanardo, line 231: 👌

lib/src/cli/flutter_cli.dart
felangel: They do unfortunately, otherwise flutter processes with lock/conflict with each other. This is an optimization we can revisit in a separate PR but I believe is out of scope for this work.

lib/src/cli/flutter_cli.dart
felangel: Why do you feel it's necessary? This is a private API and by providing a typedef developers would have to jump to the typedef to understand the function signature.

lib/src/cli/flutter_cli.dart
felangel: Same question here about whether you feel this is necessary? I'd prefer to leave it inline unless it's a public API (related to https://dart-lang.github.io/linter/lints/avoid_private_typedef_functions.html).

lib/src/cli/flutter_cli.dart
erickzanardo: Just for the sake of DRY, I've noticed this is repeated along the file.

Either way this is just a nit, I am fine either way

felangel: Where else do you see it repeated? I only see `Future<T> Function(String cwd)` in one spot.

lib/src/cli/flutter_cli.dart
erickzanardo: Interesting lint rule, never thought about it that way, makes sense! Let's keep online, we can ignore all the similar comments

lib/src/cli/flutter_cli.dart
erickzanardo: Yeah, this isn't repeated, I guess I got into the flow because of the other comments

erickzanardo (approved):
LGTM 🚀

lib/src/cli/flutter_cli.dart
felangel: The only reason I slightly prefer to leave it as is is because let's say we create a typedef like:

```dart
typedef ArgumentCallback<T> = void Function(T argument);
```

Then the new signature will look like:

```dart
Future<void> _flutterTest({
  String cwd = '.',
  required ArgumentCallback<String>? stdout,
  required ArgumentCallback<String>? stderr,
}) {...}
```

I don't really see the benefit of the typedef in this case because as a developer, I have to look at what `ArgumentCallback` means above and, from a readability standpoint, I personally prefer seeing the type inline when it's simple like this especially if it's used in a private method.

Let me know what you think 👍

erickzanardo: Got it, I think it makes sense, I kind find the inline function type a little bit confusing on dart, hence why I prefer to use typedef, and as a bonus we DRY.

But I agree with you, being a private method, we have have freedom to leave like this, and it is easier to spot the require method contract. Lets keep like that 👍

erickzanardo (approved):

Issue #279, opened 2022-02-08
feat: add support for `test` command


felangel:
As a developer, I want to be able to easily run tests and collect coverage for my projects so that I can identify failing tests and gaps in code coverage.

```sh
# run tests for current project (flutter/dart)
very_good test

# run tests recursively
very_good test --recursive

# run tests for specific project
very_good test ./path/to/my/project

# run tests with coverage generates coverage report at coverage/lcov.info
very_good test --coverage

# run tests with coverage generates coverage report at custom/path/lcov.info
very_good test --coverage --coverage-path ./custom/path

# run tests with coverage and exclude files from coverage
very_good test --coverage --coverage-exclude "*.g.dart"

# run tests with coverage and exit(1) if coverage is below min coverage threshold
very_good test --coverage --min-coverage 100

# run tests with coverage and output coverage results as html report
very_good test --coverage --report html

# run tests with coverage and output coverage results as json output
very_good test --coverage --report json

# run tests without running "pub get" first
very_good test --no-pub

# run tests in a random order
very_good test --test-randomize-ordering-seed random

```

orestesgaolin:
In the context of the https://github.com/VGVentures/employees/discussions/234 I would request for bringing improvement suggested by @jorgecoca to `very_good_cli` along with the `test` command. Long test duration with coverage is one of the major pains across multiple projects and we would be super excited to have a standardized solution to this problem.

felangel:
> In the context of the [VGVentures/employees#234](https://github.com/VGVentures/employees/discussions/234) I would request for bringing improvement suggested by @jorgecoca to `very_good_cli` along with the `test` command. Long test duration with coverage is one of the major pains across multiple projects and we would be super excited to have a standardized solution to this problem.

Yup definitely agree that internal optimizations for test performance should be considered when implementing this 👍

definitelyokay:
Linking to https://github.com/VeryGoodOpenSource/very_good_coverage/issues/20 as it seems relevant to this discussion

mtwichel:
Also adding https://pub.dev/packages/coverde to the discussion - it looks like it supports a lot of the html report generation that this aims to achive, so it may be helpful either as a dependency (although it looks like it would need to expose a lot of its interal logic for that) or just inspiration.

marcossevilla:
[...]

PR #318, opened 2022-03-17
feat(test): performance optimizations

felangel:
## Description

- feat(test): performance optimizations (part of #279)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/commands/test/test.dart
+++ b/lib/src/commands/test/test.dart
@@ -11,7 +11,9 @@ import 'package:very_good_cli/src/cli/cli.dart';
 /// {@endtemplate}
 class TestCommand extends Command<int> {
   /// {@macro test_command}
-  TestCommand({Logger? logger}) : _logger = logger ?? Logger() {
+  TestCommand({
+    Logger? logger,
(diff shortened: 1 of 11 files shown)


lib/src/commands/test/test.dart
```diff
@@ -56,21 +81,33 @@ class TestCommand extends Command<int> {
 
   @override
   Future<int> run() async {
-    final recursive = _argResults['recursive'] as bool;
     final targetPath = path.normalize(Directory.current.absolute.path);
+    final pubspec = File(path.join(targetPath, 'pubspec.yaml'));
+
+    if (!pubspec.existsSync()) {
```
erickzanardo, line 87: I m curious why moving this up and not using the exception anymore

erickzanardo (approved):
One question, but LGTM already

lib/src/commands/test/test.dart
felangel: I wanted to catch this error before running the pre gen hook since it would fail if a pubspec didn't exist.

PR #369, opened 2022-04-26
Fixed command in create new Flutter app with custom organization example

manuelperez96:
## Description

I notice that in Readme, commands section, you wrote `--org-name`, but in the following example is written like `--org`. I think that it can be a typing mistake.

## Type of Change

- [ ] 🧹 Code refactor
- [X ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Hi @manuelperez96 👋 
Thanks for opening a PR! This is working as expected though because `--org` is an alias and can be used in addition to `--org-name`

PR #379, opened 2022-05-06
feat(packages): ensure git dependencies are reachable

### LINKED ISSUE #372: fix: very_good packages get gets stuck when using dependency via git https link but not authenticated
**Description**
When using pubspec.yaml dependency via https git url, but by default ssh authentication is used (so no ability to clone repository via https), the cli hangs. This caused us some problems as the CI was stuck for many hours on the dependency resolving step.

**Steps To Reproduce**

0. Use ssh instead of https authentication on GitHub
1. Add pubspec.yaml dependency using https link to a private GitHub repository e.g.   
```yaml 
my_package:
    git:
      url: https://github.com/org/package.git
      ref: 1a61e2a62094aaf7614a4f8a0063863d21cc6e44
```
2. Run `very_good packages get`
3. See following prompt indefinitely

```sh
❯ very_good packages get
⠋ Running "flutter packages get" in⠸ Running "flutter packages get" in /Users/...
```

**Expected Behavior**
The action should fail with the clear information

**Additional Context**

Maybe timeout would be useful?

felangel:
## Description

- feat(packages): ensure git dependencies are reachable (closes #372)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/test/src/cli/git_cli_test.dart
+++ b/test/src/cli/git_cli_test.dart
@@ -1,0 +1,27 @@
+import 'package:test/test.dart';
+import 'package:very_good_cli/src/cli/cli.dart';
+
+void main() {
+  group('Git', () {
+    group('reachable', () {
+      test('completes for a reachable remote', () async {
+        await expectLater(
+          Git.reachable(
+            Uri.parse('https://github.com/verygoodopensource/very_good_cli'),
+          ),
+          completes,
+        );
+      });
+
+      test('throws UnreachableGitDependency for an unreachable remote',
+          () async {
+        await expectLater(
+          Git.reachable(
+            Uri.parse('https://github.com/verygoodopensource/_very_good_cli'),
+          ),
+          throwsA(isA<UnreachableGitDependency>()),
+        );
+      });
+    });
+  });
+}
--- a/test/src/cli/git_cli_test.dart
+++ b/test/src/cli/git_cli_test.dart
@@ -23,5 +23,21 @@ void main() {
         );
       });
     });
+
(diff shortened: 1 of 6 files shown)


test/src/cli/git_cli_test.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:test/test.dart';
+import 'package:very_good_cli/src/cli/cli.dart';
+
+void main() {
+  group('Git', () {
+    group('reachable', () {
+      test('completes for a reachable remote', () async {
+        await expectLater(
+          Git.reachable(
```
wolfenrain, line 9: This test would fail if you don't have GIT on your system or in the CI right? While that would be strange to not have, it would fail incorrectly cause the test itself would probably be fine.

Should we maybe mock the _Cmd somehow?

wolfenrain (approved):
One small nit, but lgtm

test/src/cli/git_cli_test.dart
felangel: Yeah I was planning to create a separate issue to refactor all CLI classes to support an injectable Process rather than exposing static methods. I'll update this handle checking if git is not installed, good point 👍

wolfenrain: True but technically that could be skipped by the user, but I am fine with improving it for the tests on a separate PR!

felangel: https://github.com/VeryGoodOpenSource/very_good_cli/issues/380

PR #388, opened 2022-05-10
fix: run golden tests with test optimization enabled

### LINKED ISSUE #363: fix: golden tests don't work with test optimisation flag
**Description**

Testing goldens don't work when using the test optimisation. It throws an error, only work around at the moment is not using optimisation.

**Steps To Reproduce**

1. Write a test with a golden in it.
2. Run `very_good test --update-goldens`
3. Run `very_good test --optimization`

**Expected Behavior**
That I can run my golden tests while utilising test optimisation.

**Additional Context**
```
Running "flutter test" in /Users/wolfen/Projects/some_project...
✓ Optimizing tests (5.0s)
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following TestFailure was thrown running a test:
Expected: one widget whose rasterized image matches golden image "golden/my_component.png"
  Actual: _WidgetPredicateFinder:<exactly one widget with widget matching predicate (Closure:
(Widget) => bool) (ignoring offstage widgets): GameWidget<TestGame>(state:
_GameWidgetState<TestGame>#11074)>
   Which: Could not be compared against non-existent file: "golden/my_component.png"

When the exception was thrown, this was the stack:
#0      fail (package:test_api/src/expect/expect.dart:137:31)
#1      _expect.<anonymous closure> (package:test_api/src/expect/expect.dart:112:9)
<asynchronous suspension>
<asynchronous suspension>
(elided 6 frames from dart:async and package:stack_trace)

The test description was:
  renders correctly
════════════════════════════════════════════════════════════════════════════════════════════════════
Test failed. See exception logs above.
The test description was: renders correctly
components_my_component_test_dart MyComponent renders correctly /Users/wolfen/Projects/some_project/test/.test_runner.dart (FAILED)
00:04 +1 -1: Some tests failed.
```

Giuspepe:
## Description
This fixes #363.

Currently, golden tests fail when running them with `very_good test` with test optimization enabled. This is because `goldenFileComparator` searches for the golden files relative from the currently running test file. With optimization enabled, the golden files are searched from the `.test_runner.dart` file instead of from the original test files.

The solution I suggest is to use a custom implementation for `goldenFileComparator` which knows about the implications of the test optimization. It finds the correct absolute path of the golden file by searching for a single file ending with the given relative path from the test base directory. For further details, see my comment here: https://github.com/VeryGoodOpenSource/very_good_cli/issues/363#issuecomment-1122548970

This approach fails in the following situation:
- test file `test/a/foo_test.dart` uses the relative path `goldens/foo.png` to refer to the golden image at `test/a/goldens/foo.png`
- test file `test/b/foo_test.dart` uses the same relative path `goldens/foo.png` to refer to the golden image at `test/b/goldens/foo.png`

In that case, searching for files ending with the path `goldens/foo.png` does not return a single result but two possible results. An exception will be thrown in that case. I think this is ok because it is a rather theoretical edge case and this approach solves all realistic cases.
 
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


Giuspepe:
@felangel would you mind reviewing this? Thanks!

wolfenrain:
Thank you for your contribution!

Our bundles come from an internal brick repository currently, so we normally supply the change there and then do a PR here. But I will merge this as-is so we don't lose your contribution, as soon as the CI succeeds!

Just for your information, I'll supply some extra changes from our internal bricks in a separate PR to keep it up to date with that repo. In case you see another PR today that changes your code ;-), it wont change any of the functionality that your provided.

Once again thanks for this PR!

PR #461, opened 2022-07-15
fix: move alert messages to custom method

wolfenrain:
## Description

After talking with @renancaraujo and @felangel we came to the conclusion that we were abusing the alert message, it is meant to indicate something severe or a call to action but we mostly used it for the pretty colors.

Fixing that by moving to a custom message that uses these pretty colors.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/logger_extension.dart
+++ b/lib/src/logger_extension.dart
@@ -1,0 +1,9 @@
+import 'package:mason_logger/mason_logger.dart';
+
+/// Extension on the Logger class for custom styled logging.
+extension LoggerX on Logger {
+  /// Log a message in the "created" style of the CLI.
+  void created(String message) {
+    info(lightCyan.wrap(styleBold.wrap(message)));
+  }
+}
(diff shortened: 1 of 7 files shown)


renancaraujo (approved):

lib/src/logger_extension.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:mason_logger/mason_logger.dart';
+
+/// Extension on the Logger class for custom styled logging.
+extension LoggerX on Logger {
+  /// Log a message in the "created" style of the CLI.
+  void created(String message) {
```
felangel, line 6: Minor but thoughts on calling this something like `pretty` or `emphasis`?

```suggestion
  void emphasis(String message) {
```

felangel (approved):

lib/src/logger_extension.dart
felangel: I’m fine either way just was thinking about if we have another non created message with the same styling

PR #473, opened 2022-07-28
feat(test): add --reporter to output in JSON

jackgllghr:
## Description

Added the `--reporter` flag in the test command, to allow the user to output the test results in JSON. This change simply passes through the command and disables logging when outputting to JSON

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


jackgllghr:
I'm not sure if it makes more sense to the use `--machine` arg or `--reporter` arg, I just tried to stick to the original flutter command. 

I'm having some issues running the tests in my local environment, some tests I haven't gone near are failing so apologies in advance if I missed something glaring

lib/src/commands/test/test.dart
```diff
@@ -57,6 +57,11 @@ class TestCommand extends Command<int> {
         defaultsTo: true,
         help: 'Whether to apply optimizations for test performance.',
       )
+      ..addFlag(
+        'machine',
```
felangel, line 61: I'd prefer to keep the options consistent with the flutter tool as much as possible. In this case rather than `--machine` can we stick to `--reporter json`?

jackgllghr: Updated it there, but I'm having trouble testing locally, from a clone of the repo I can't get it building/testing successfully. I'm using `dart run bin/very_good.dart test -h` but when it comes to running the tests for this repo, it's not succeeding. Any ideas? Dependencies installed and all, my flutter is 3.0.5, anything else it could be missing?

felangel: I don't think these changes are enough unfortunately. We're already using the flag internally but we would need to adjust the output based on whether the user explicitly requested a custom reporter.

Thelm76:
Hi! any news on this PR? I would also need it to get a JSON report for my pipeline....

Thelm76:
@felangel, @jackgllghr is this PR dead? Sorry if I'm beeing to insistent...

felangel:
> @felangel, @jackgllghr is this PR dead? Sorry if I'm beeing to insistent...

It's not dead but the changes included in this PR are not sufficient so additional work needs to be done to get this to a mergable state.

jackgllghr:
My apologies, the project I required this for got cancelled so my focus wavered. I might get a chance to look at this over the Xmas break, but if someone wants to jump in and finish it off, go for it

felangel:
> My apologies, the project I required this for got cancelled so my focus wavered. I might get a chance to look at this over the Xmas break, but if someone wants to jump in and finish it off, go for it

No problem at all! I'm going to close the PR for now but feel free to reopen it if you have time to revisit this, thanks!

PR #492, opened 2022-09-12
feat(test): add --dart-define

Giuspepe:
## Description
This PR adds the `--dart-define` parameter to the test command. 

`--dart-define` is used to define compile-time constants which are then available through the `String.fromEnvironment`, `bool.fromEnvironment`, `int.fromEnvironment`, and `double.fromEnvironment` constructors. 
Tests relying on such compile-time constants can currently not be run with `very_good test`, because the command doesn't yet accept the `--dart-define` parameter.

An example where one would use a `--dart-define` compile-time constant is to switch between different environments or flavors of an app. E.g. `--dart-define=APP_ENVIRONMENT=dev`. 
This is useful because code as the following is automatically deleted in an optimization step when building an app:
```
const environment = String.fromEnvironment('APP_ENVIRONMENT');
const isEnvironmentDev = environment == 'dev';

...

// the whole if-block is automatically deleted during optimization when APP_FLAVOR != dev
// because `isEnvironmentDev` is a compile-time constant
if(isEnvironmentDev) {
  final secretDataOnlyForDev = 42;
}
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


Giuspepe:
FYI: It seems that the CI test task is a bit flaky. I got the following error on my first CI run. When I reran the CI task with the same code, it succeeded without any errors.

https://github.com/VeryGoodOpenSource/very_good_cli/runs/8308620438?check_suite_focus=true
image

felangel (approved):
LGTM thanks so much for the contribution!

PR #522, opened 2022-10-19
feat(test): support @Tags annotation

mugbug:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/commands/test/templates/test_runner_bundle.dart
```diff
@@ -1,5 +1,5 @@
 // GENERATED CODE - DO NOT MODIFY BY HAND
-// ignore_for_file: prefer_single_quotes, public_member_api_docs, lines_longer_than_80_chars, implicit_dynamic_list_literal, implicit_dynamic_map_literal
+// ignore_for_file: type=lint, implicit_dynamic_list_literal, implicit_dynamic_map_literal, inference_failure_on_collection_literal
```
mugbug, line 2: I don't have access to the original `test_runner` brick repo, so I couldn't open a PR there. Also, I suppose some tests should be added to the pregen hook, but I was in doubt either if the `unbundle` command also unbundles test files from a brick and some tips on best practices for testing bricks

mugbug:
> Hello Mr @mugbug, sorry for the huge delay,

Hey @renancaraujo! No problem, thanks for the reply and detailed response! And indeed, I haven't considered many of those extra features related to tagging the tests. And for sure it seems a lot more effort for a minimum outcome since I feel like not a lot of people would use/need it. Also, my use case is a temporary one. We just needed this feature to enable the usage of `very_good test` right away without any refactoring, and it's doing its job well, but in a near future we just want to remove all file tags so all tests run with optimization.

That said, to me it also doesn't make much sense to consider all those topics you mentioned and increment this change, so I'm ok on just closing this PR and keep installing very_good_cli through git ref in a while until we get to run all tests with it.

PR #640, opened 2023-02-06
feat: add support for excluding packages

### LINKED ISSUE #636: CLI - feat: support excluding directories
As a developer, I may have directories which I wish to exclude from 'packages get' and Test when using the recursive flag.

- [ ] Packages Get
- [ ] Test

willhlas:
## Description

Adds support for excluding packages when running `packages get`.

- `very_good packages get -r --ignore=pluginA,pluginB`

Closes #636 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/cli/cli.dart
```diff
@@ -140,9 +140,10 @@ const _ignoredDirectories = {
   '.fvm',
 };
 
-bool _isPubspec(FileSystemEntity entity) {
+bool _isPubspec(FileSystemEntity entity, Set<String> ignore) {
   final segments = p.split(entity.path).toSet();
   if (segments.intersection(_ignoredDirectories).isNotEmpty) return false;
+  if (segments.intersection(ignore).isNotEmpty) return false;
```
felangel, line 146: I don't think this will work if we pass a glob which I believe should be supported since most developers will likely expect globs to be supported.

```
flutter packages get -r --ignore "bricks/**"
```

lib/src/cli/cli.dart
```diff
@@ -140,9 +140,10 @@ const _ignoredDirectories = {
   '.fvm',
 };
 
-bool _isPubspec(FileSystemEntity entity) {
+bool _isPubspec(FileSystemEntity entity, Set<String> ignore) {
```
felangel, line 143: I don't think we should pass the ignore list into this function since this function should be specifically determining whether a `FileSystemEntity` is a pubspec.

willhlas: What do you think of having bool checks like this:
```dart
bool _isPubspec(FileSystemEntity entity) {
  if (entity is! File) return false;
  return p.basename(entity.path) == 'pubspec.yaml';
}

bool _isIgnored(FileSystemEntity entity, Set<String> ignore) {
  final segments = p.split(entity.path).toSet();
  if (segments.intersection(_ignoredDirectories).isNotEmpty) return true;
  if (segments.intersection(ignore).isNotEmpty) return true;
  return false;
}
```

and ran with:
```dart
final processes = _Cmd.runWhere<T>(
  run: (entity) => cmd(entity.parent.path),
  where: (entity) => !_isIgnored(entity, ignore) && _isPubspec(entity),
  cwd: cwd,
);
```

felangel: Yeah that looks better to me. We could even add a simple extension method:

```dart
final processes = _Cmd.runWhere<T>(
  run: (entity) => cmd(entity.parent.path),
  where: (entity) => ignore.excludes(entity) && _isPubspec(entity),
  cwd: cwd,
);
```

lib/src/cli/cli.dart
willhlas: What do you think about this?
```dart
extension on Set<String> {
  bool excludes(FileSystemEntity entity) {
    final segments = p.split(entity.path).toSet();
    if (segments.intersection(_ignoredDirectories).isNotEmpty) return true;
    if (segments.intersection(this).isNotEmpty) return true;

    for (final value in this) {
      if (Glob(value).matches(entity.path)) {
        return true;
      }
    }

    return false;
  }
}
```

willhlas: Potentially could split this up if you think it's better 👍🏻 
```dart
for (final value in this) {
  final glob = Glob(value);
  if (glob.matches(entity.path)) {
    return true;
  }
}
```

erickzanardo (dismissed):

wolfenrain (approved):

renancaraujo (dismissed):

felangel (changes requested):
Can we add a test to ensure globs are supported?

lib/src/cli/cli.dart
felangel: Yeah that seems reasonable to me

lib/src/cli/cli.dart
```diff
@@ -143,8 +143,23 @@ const _ignoredDirectories = {
...
 bool _isPubspec(FileSystemEntity entity) {
-  final segments = p.split(entity.path).toSet();
-  if (segments.intersection(_ignoredDirectories).isNotEmpty) return false;
   if (entity is! File) return false;
   return p.basename(entity.path) == 'pubspec.yaml';
 }
+
+extension on Set<String> {
+  bool excludes(FileSystemEntity entity) {
+    final segments = p.split(entity.path).toSet();
```
felangel, line 152: Take the following case:

`ignore: {'dev/projects/dev/'}`
`entity.path`: `dev/projects/dev/foo`

Segments would equal `{'dev', 'projects', 'foo'}` which means the intersection of `segments` and `ignoredDirectories` would be empty and the intersection of `segments` and `this` would also be empty.

willhlas: Setting `recursive` to `true` and running `very_good packages get --recursive --ignore "dev/projects/dev"` completes as expected.

```dart
for (final value in this) {
  final glob = Glob(value, recursive: true);
  if (glob.matches(entity.path)) {
    return true;
  }
}
```

From the `Glob` docs:
```
If [recursive] is true, this glob matches and lists not only the files and directories it explicitly matches, but anything beneath those as well.
```

Let me know what you think!

willhlas: Never mind actually, I forgot to add `/` to the end of `"dev/projects/dev"`.

renancaraujo (approved):

test/src/commands/packages_test.dart
```diff
@@ -265,6 +266,64 @@ void main() {
...
+            [
+              'packages',
+              'get',
+              '--recursive',
+              '--ignore=plugin_b',
+              tempDirectory.path,
+            ],
+          );
+          expect(result, equals(ExitCode.success.code));
+          verify(() {
```
felangel, line 320: Shouldn't we verify that we're running flutter packages get in the specific packages? The current verify isn't very specific and this test would likely pass even without the changes in the branch.

test/src/cli/flutter_cli_test.dart
```diff
@@ -168,25 +168,62 @@ void main() {
...
+              completes,
             ),
-            completes,
-          ),
-          runProcess: process.run,
-        );
-      });
+            runProcess: process.run,
+          ).whenComplete(() {
+            verify(() {
```
felangel, line 219: Shouldn't we verify that we're running flutter packages get in the specific packages? The current verify isn't very specific and this test would likely pass even without the changes in the branch.

felangel (approved):

lib/src/cli/cli.dart
```diff
@@ -162,8 +162,31 @@ const _ignoredDirectories = {
...
+          value = '$value**';
+        } else {
+          value = '$value/**';
+        }
+      }
+
+      final glob = Glob(value);
+      if (glob.matches(entity.path)) {
+        return true;
+      }
```
wolfenrain, line 187: Why are we turning the `value` into a glob when there is a `/` but no `*` in the string? Could you elaborate on what this is doing exactly?

willhlas: For some reason if the `value` is  `dev/projects/dev` (example), `Glob(value)` wouldn't find a match. My workaround is to add the glob pattern based on the `value` so that `dev/projects/dev` and `dev/projects/dev/` would both convert to `dev/projects/dev/**` so that the glob can find a match!

wolfenrain: Hmm but if that's the case I think we can get away by doing it like this:

```suggestion
      if (value.contains('*') {
         return Glob(value).matches(entity.path);
      }
      return entity.path.startsWith(value);
```

willhlas: Tricky part is that `entity.path.startsWith(value)` would always return false since the ignored package (value), isn't the full path like `Users/user/my_project/packages/dev/...`. 

So if I ran `very_good packages get --recursive --ignore packages/dev`, this would still get packages in `packages/dev`

Technically using `entity.path.contains(value)` will work, but not sure if this would be ideal.

wdyt?

wolfenrain: Just synced with @willhlas on this and the use case of a string like `dev/projects/dev` without a `*` is not a valid glob according to the specs. So if we are only supporting the glob pattern we should not introduce non-glob patterns for matching purposes.

So let's revert to the original implementation of only doing a `return Glob(value).matches(entity.path)`

lib/src/cli/cli.dart
```diff
@@ -162,8 +162,20 @@ const _ignoredDirectories = {
...
+
+extension on Set<String> {
+  bool excludes(FileSystemEntity entity) {
+    final segments = p.split(entity.path).toSet();
+    if (segments.intersection(_ignoredDirectories).isNotEmpty) return true;
+    if (segments.intersection(this).isNotEmpty) return true;
+
+    for (final value in this) {
+      return Glob(value).matches(entity.path);
+    }
```
wolfenrain, line 177:
```suggestion
    for (final value in this) {
      if (Glob(value).matches(entity.path)) {
        return true;
      }
    }
```

Otherwise it will end early right?

willhlas: Ah yeah good catch!

wolfenrain (approved):

renancaraujo (approved):

PR #658, opened 2023-02-20
fix: test optimizer to import files as valid dart names

### LINKED ISSUE #648: fix: very_good test doesn't like [ in file names
**Description**
very_good test fails when files have [] in their filenames (as seen in [Dart Frog dynamic route](https://dartfrog.vgv.dev/docs/basics/routes#dynamic-routes-) files).

**Steps To Reproduce**

**Expected Behavior**

**Screenshots**

image

**Additional Context**

alestiago:
## Description

Instead of importing files as a snaked case name of the relative path, the test files are assigned an alphabetic identifier.

Before:
```dart
import 'something[1]_test.dart' as something[1]_test; // Invalid dart identifier
```

After:
```dart
import 'something[1]_test.dart' as _a;
```

Resolves #648 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -28,21 +28,55 @@ Future<void> run(HookContext context) async {
...
 extension on FileSystemEntity {
   bool get isTest {
     return this is File && path.basename(this.path).endsWith('_test.dart');
   }
 }
+
+/// {@template string_id_generator}
+/// A class that generates short identifiers.
+/// {@endtemplate}
+class StringIdGenerator {
```
alestiago, line 55: This algorithm was given by @wolfenrain .

If we want to adjust the logic of how identifier names are generated I'm open to changes.

bricks/test_optimizer/brick.yaml
```diff
@@ -1,6 +1,6 @@
 name: test_optimizer
 description: A brick that generates a single entrypoint for Dart tests.
-version: 0.1.0+1
+version: 0.1.0+2
```
wolfenrain, line 3: We dont publish this brick, maybe we should even remove the version.

/cc @renancaraujo @felangel

bricks/test_optimizer/__brick__/test/.test_optimizer.dart
```diff
@@ -6,11 +6,11 @@
 import 'package:flutter_test/flutter_test.dart';
 {{/isFlutter}}{{^isFlutter}}import 'package:test/test.dart';{{/isFlutter}}
 
-{{#tests}}import '{{{.}}}' as {{#snakeCase}}{{{.}}}{{/snakeCase}};
+{{#tests}}import '{{{path}}}' as {{identifier}};
 {{/tests}}
 void main() {
 {{#isFlutter}}  goldenFileComparator = _TestOptimizationAwareGoldenFileComparator();{{/isFlutter}}
-{{#tests}}  group('{{#snakeCase}}{{{.}}}{{/snakeCase}}', () { {{#snakeCase}}{{{.}}}{{/snakeCase}}.main(); });
+{{#tests}}  group('{{identifier}}', () { {{identifier}}.main(); });
```
wolfenrain, line 13:
```suggestion
{{#tests}}  group('{{path.snakeCase()}}', () { {{identifier}}.main(); });
```

Let's keep the group names the same as before, that way the information conveyed in case of failure is more clearer?

bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -28,21 +28,55 @@ Future<void> run(HookContext context) async {
...
-      .map(
-        (entity) => path
-            .relative(entity.path, from: testDir.path)
-            .replaceAll(r'\', '/'),
-      )
-      .toList();
-
-  context.vars = {'tests': tests, 'isFlutter': isFlutter};
+  final identifierGenerator = StringIdGenerator();
+  final testIdentifierTable = <Map<String, String>>[];
```
wolfenrain, line 32: nit: I would rename this to `testsNormalized`, something a bit descriptive on the why

bricks/test_optimizer/hooks/lib/pre_gen.dart
wolfenrain, line 55: We can put this class in it's own file now with the newer version of Mason (cc @felangel)

bricks/test_optimizer/hooks/test/string_id_generator_test.dart
```diff
@@ -0,0 +1,29 @@
...
+      expect(StringIdGenerator.new, returnsNormally);
+    });
+
+    group('next', () {
+      test('returns normally', () {
+        final generator = StringIdGenerator();
+        expect(generator.next, returnsNormally);
+      });
+
+      test('generates unique strings', () {
```
wolfenrain, line 16: We should also test the expect order of the generated, what would happen after the last index was z? Will it be `zA` or `aa`?

alestiago: Do we care? My only concern is that they are unique.

wolfenrain: To ensure a certain uniqueness, yes ;-)

That is the goal of this class and the unit. Let's say this class will be used for something that is long lived, if this isn't tested and we change it we might accidentally introduce a breaking change because our output of the CLI would change.

bricks/test_optimizer/hooks/lib/pre_gen.dart
alestiago: Side note:

Going to alter the algorithm to be a1, a2, a3, and so on. Since the current algorithm can generate words that are dart keywords which can't be used as indentifiers.

bricks/test_optimizer/brick.yaml
alestiago: I'll set it back to 0.1.0+1 since removing violates the brick schema. Wdyt?

bricks/test_optimizer/hooks/lib/pre_gen.dart
alestiago: I think testIdentifierTable is descriptive. Reading as "matches a test file to its dart identifier".

I like to include the "table" in the variable to resemble common computer science terms such as finger tables as in a Distributed Hash Table (DHT), or symbol tables in the compile process.

I updated the README example to give some intuition on why this is done.

bricks/test_optimizer/__brick__/test/.test_optimizer.dart
```diff
@@ -6,11 +6,11 @@
 import 'package:flutter_test/flutter_test.dart';
 {{/isFlutter}}{{^isFlutter}}import 'package:test/test.dart';{{/isFlutter}}
 
-{{#tests}}import '{{{.}}}' as {{#snakeCase}}{{{.}}}{{/snakeCase}};
+{{#tests}}import '{{{path}}}' as {{identifier}};
 {{/tests}}
 void main() {
 {{#isFlutter}}  goldenFileComparator = _TestOptimizationAwareGoldenFileComparator();{{/isFlutter}}
-{{#tests}}  group('{{#snakeCase}}{{{.}}}{{/snakeCase}}', () { {{#snakeCase}}{{{.}}}{{/snakeCase}}.main(); });
+{{#tests}}  group('{{path.snakeCase()}}', () { {{identifier}}.main(); });
```
alestiago, line 13: @wolfenrain wdyt of removing the .snakeCase here?

So the group has the file name:
```dart
group('app/view/app_test.dart', () {})
// Instead of 
group('app_view_app_test_dart', () {})
```

wolfenrain: That would fail if the path contains a `'` which is a technically possible path. I think that is why we originally did the snakeCase as that removes all of those issues

bricks/test_optimizer/hooks/lib/dart_identifier_generator.dart
```diff
@@ -0,0 +1,35 @@
+/// {@template DartIdentifierGenerator}
```
wolfenrain, line 1:
```suggestion
/// {@template dart_identifier_generator}
```

bricks/test_optimizer/hooks/lib/dart_identifier_generator.dart
```diff
@@ -0,0 +1,35 @@
...
+/// A class that generates valid Dart identifiers.
+///
+/// See also:
+///
+/// * Section 17.37 from [Dart Language Specification](https://dart.dev/guides/language/specifications/DartLangSpec-v2.10.pdf)
+/// {@endtemplate}
+class DartIdentifierGenerator {
+  /// {@macro DartIdentifierGenerator}
```
wolfenrain, line 9:
```suggestion
  /// {@macro dart_identifier_generator}
```

wolfenrain (approved):

bricks/test_optimizer/__brick__/test/.test_optimizer.dart
alestiago: Awesome, I'll keep it as is and update the README.

renancaraujo (approved):

wolfenrain (approved):

PR #704, opened 2023-03-28
fix: support ignoring empty globs

### LINKED ISSUE #703: fix: allow specifying an empty string as a Glob
**Description**
When using Very Good CLI 0.12.3 doing:

```sh
very_good packages get -r --ignore=""
```

Shows:
```sh
Error on line 1, column 1: expected a glob.
  ╷
1 │ 
  │ ^
  ╵
```

**Steps To Reproduce**

1. Run `very_good packages get -r --ignore=""` in a valid directory

**Expected Behavior**
To accept an empty string as input.

**Additional Context**
Would simplify [feat/support-excluding-directories](https://github.com/VeryGoodOpenSource/very_good_workflows/pull/110) default parameter.

alestiago:
## Description

Resolves #703 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/cli/cli.dart
```diff
@@ -176,9 +176,7 @@ extension _Set on Set<String> {
     if (segments.intersection(this).isNotEmpty) return true;
 
     for (final value in this) {
-      if (Glob(value).matches(entity.path)) {
-        return true;
-      }
+      return value.isNotEmpty && Glob(value).matches(entity.path);
     }
```
wolfenrain, line 180: This wont ever loop over multiple values, as it will return on the first loop. That is why we had the `if` statement there.

alestiago: Indeed, thanks for noticing it! Completely forgot space-separated globs are accepted.

wolfenrain (approved):

PR #835, opened 2023-10-04
ci: include workflow to commit SPDX updates as bot

alestiago:
## Description

Creates a workflow that will trigger a bot to automatically update the SPDX License enumeration whenever the list is changed.

## Type of Change

- [ ] 🧹 Code refactor
- [X] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


.github/workflows/spdx_license_bot.yaml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          base: main
+          branch: chore/update-spdx-license
+          commit-message: "chore: update SPDX licenses"
+          title: "chore: update SPDX licenses"
+          body: Please squash and merge me!
+          labels: bot
+          author: VGV Bot <vgvbot@users.noreply.github.com>
+          assignees: vgvbot
+          reviewers: alestiago
```
wolfenrain, line 63: I would keep this empty and just have the CODEOWNER file trigger

.github/workflows/spdx_license_bot.yaml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: 📚 Git Checkout
+        uses: actions/checkout@v4
+
+      - name: 🎯 Setup Dart
+        uses: dart-lang/setup-dart@v1
+        with:
+          sdk: 3.1.0
```
wolfenrain, line 31: I would just go with stable and it will break eventually, that way we always use whatever is stable instead of old :P

.github/workflows/spdx_license_bot.yaml
alestiago: Awesome, I will also edit https://github.com/VeryGoodOpenSource/very_good_cli/blob/2a7ad1e0d3e2a471036b0ecf0bffd91cfcd99aa3/.github/workflows/bump_templates.yaml#L39.

erickzanardo (approved):

erickzanardo (approved):

PR #841, opened 2023-10-05
feat: include hidden check licenses command

alestiago:
## Description

Adds `very_good packages check` and `very_good packages check licenses` as hidden commands. This is to anticipate #771 .

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/commands/packages/commands/check/check.dart
```diff
@@ -0,0 +1,24 @@
...
+class PackagesCheckCommand extends Command<int> {
+  /// {@macro packages_check_command}
+  PackagesCheckCommand({Logger? logger}) : _logger = logger ?? Logger() {
+    addSubcommand(PackagesCheckLicensesCommand(logger: _logger));
+  }
+
+  final Logger _logger;
+
+  @override
+  String get description => 'Check packages in a Dart or Flutter project.';
```
erickzanardo, line 17: Checks what? I believe we could expand this description a bit more in order to make it clearer what kind of things are verified

alestiago: As of this PR, this command does nothing at all, similarly to doing `very_good packages`. Doing `very_good packages check` will output what checks you can do (same behaviour as `very_good packages`).

erickzanardo: Yeah, I noticed that as I went through the PR, but we already have an idea on what it will do right? I still think we should elaborate more in its description in this PR, since that description will set the tone for follow ups.

renancaraujo (approved):

PR #846, opened 2023-10-10
feat: add `--ignore-failures` to `check licenses`

alestiago:
> Should be merged after #842 

## Description

Adds a flag (`--ignore-failures`) on `very_good packages check licenses`.

This change anticipates https://github.com/VeryGoodOpenSource/very_good_cli/issues/771

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #847, opened 2023-10-10
feat: add `--dependency-type` to `check licenses`

alestiago:
> Should be merged after #846 

## Description

Adds a multi option (--dependecy-type) on very_good packages check licenses.

This change anticipates https://github.com/VeryGoodOpenSource/very_good_cli/issues/771

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/commands/packages/commands/check/commands/licenses.dart
```diff
@@ -22,11 +22,27 @@ class PackagesCheckLicensesCommand extends Command<int> {
...
+        negatable: false,
+      )
+      ..addMultiOption(
+        'dependency-type',
+        help: 'The type of dependencies to check licenses for.',
+        allowed: [
+          'direct-main',
+          'direct-dev',
+          'transitive',
+        ],
```
wolfenrain, line 38: Do we want an `all`?

alestiago: Is a multi-option so for all you should do: `--dependency-type direct-main,direct-dev,transitive`. I think all might be adding unnecessary additional complexity? Might lead to additional questions from the user: "why can I use _all_ and _direct-main_"?

wolfenrain (approved):

PR #869, opened 2023-10-23
fix: avoid failure when terminal has no columns

### LINKED ISSUE #866: fix: StdoutException: Could not get terminal size, OS Error: Operation not supported on socket, errno = 102
**Description**
I'm using the `very_good test --coverage` command in a pipeline in Azure Devops, and it's returning this error when trying to write the message in the terminal:

"Thank you for using Very Good Ventures open source tools!
Don't forget to fill out this form to get information on future updates and releases here:
https://verygood.ventures/open-source/cli/subscribe-latest-tool-updates"

**Steps To Reproduce**

1. Install lastest version 1.17.0
2. Run `very_good test --coverage`

**Expected Behavior**

**Screenshots**
image

**Additional Context**
Locally works normally

alestiago:
## Description

Resolves #866 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/logger_extension.dart
+++ b/lib/src/logger_extension.dart
@@ -1,3 +1,4 @@
+import 'dart:io';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:meta/meta.dart';
 import 'package:universal_io/io.dart';
@@ -31,10 +32,11 @@ extension LoggerX on Logger {
     int? length,
   }) {
     late final int maxLength;
-    try {
-      maxLength = length ?? stdout.terminalColumns;
-    } on StdoutException catch (_) {
-      // Not all terminals have a width, so we fallback to a reasonable value.
+    if (length != null) {
+      maxLength = length;
+    } else if (stdout.hasTerminal) {
+      maxLength = stdout.terminalColumns;
+    } else {
       maxLength = fallbackStdoutTerminalColumns;
     }
 
--- a/lib/src/logger_extension.dart
+++ b/lib/src/logger_extension.dart
@@ -20,9 +20,11 @@ extension LoggerX on Logger {
   /// Wrap the [text] to fit perfectly within the width of the terminal when
   /// [print]ed.
   ///
-  /// To overwrite the width you can use [length]; otherwise, it defaults to the
-  /// width of the terminal, or to [defaultStdoutTerminalColumns] if the width
-  /// cannot be determined.
+  /// The text will wrap around the terminal width, and will not break words. If
+  /// the terminal width cannot be determined, the text will wrap around the
+  /// [defaultStdoutTerminalColumns].
+  ///
+  /// To completely overwrite the width you can use [length].
   void wrap(
     String? text, {
     required void Function(String?) print,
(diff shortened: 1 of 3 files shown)


erickzanardo (dismissed):

lib/src/logger_extension.dart
```diff
@@ -11,13 +20,24 @@ extension LoggerX on Logger {
...
     String? text, {
     required void Function(String?) print,
     int? length,
   }) {
-    final maxLength = length ?? stdout.terminalColumns;
+    late final int maxLength;
+    try {
+      maxLength = length ?? stdout.terminalColumns;
+    } on StdoutException catch (_) {
+      // Not all terminals have a width, so we fallback to a reasonable value.
```
wolfenrain, line 37: It is only thrown if the terminal has no stdout, instead we should just check for [hasTerminal](https://api.dart.dev/stable/2.13.3/dart-io/Stdout/hasTerminal.html) and use that as the fallback.

alestiago: I amended the code. For other readers this information is documented within the `terminalColumns` method:

> If no terminal is attached to stdout, a [StdoutException] is thrown. See [hasTerminal] for more info.

lib/src/logger_extension.dart
```diff
@@ -11,13 +21,25 @@ extension LoggerX on Logger {
...
   }) {
-    final maxLength = length ?? stdout.terminalColumns;
+    late final int maxLength;
+    if (length != null) {
+      maxLength = length;
+    } else if (stdout.hasTerminal) {
+      maxLength = stdout.terminalColumns;
+    } else {
+      maxLength = fallbackStdoutTerminalColumns;
+    }
```
wolfenrain, line 41:
```suggestion
    final maxLength = switch (length) {
      == null when stdout.hasTerminal => stdout.terminalColumns,
      == null when !stdout.hasTerminal => fallbackStdoutTerminalColumns,
      _ => length
    };
```

alestiago: I think this is Nit. Tbh, I think the suggestion is less legible. Does it have an impact on performance?

wolfenrain: Then u add a `!` to it

wolfenrain: it is guaranteed to be non-null anyway, but because `hasTerminal` is not a constant value it can't figure that out

wolfenrain (approved):

lib/src/logger_extension.dart
```diff
@@ -11,13 +21,22 @@ extension LoggerX on Logger {
...
+  ///
+  /// To completely overwrite the width you can use [length].
   void wrap(
     String? text, {
     required void Function(String?) print,
     int? length,
   }) {
-    final maxLength = length ?? stdout.terminalColumns;
+    final maxLength = switch (length) {
+      == null when stdout.hasTerminal => stdout.terminalColumns,
```
erickzanardo, line 35: Oh, this is new to me, pretty cool

erickzanardo (approved):

Issue #754, opened 2023-06-19
fix: package name not visible in output when run tests recursive


DFelten:
**Description**

When running all tests recursive the package title is not visible in the output. 

`very_good test --optimization --coverage -r`

Folder structure:

```
packages/name123
packages/name456
packages/name789
```

Output:

```
✓ Optimizing tests (56ms)
00:16 +58: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages...
✓ Optimizing tests (61ms)
00:20 +377: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages...
✓ Optimizing tests (54ms)
00:14 +31: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages...
✓ Optimizing tests (57ms)
```

**Expected Behavior**

Title of each package is visible in the output:

```
✓ Optimizing tests (56ms)
00:16 +58: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages/name123...
✓ Optimizing tests (61ms)
00:20 +377: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages/name456...
✓ Optimizing tests (54ms)
00:14 +31: All tests passed!
Running "flutter test" in /Users/danielfelten/Projekte/test/packages/name789...
✓ Optimizing tests (57ms)
```

superiorsd10:
Hello @DFelten 👋, 
I wanted to test if my changes to the fork for this issue are correct or not. How can I test it by replicating the project structure provided by you?

DFelten:
> Hello @DFelten 👋, I wanted to test if my changes to the fork for this issue are correct or not. How can I test it by replicating the project structure provided by you?

Hi @superiorsd10, I'm glad you want to implement the feature. The folder structure is described above. You just need to create several packages and run the tests in the root directory.

PR #920, opened 2023-12-13
refactor: Use relative path instead of absolute when running `test`

mrgnhnt96:
Fixes: #754

## Status

**READY**

## Description

The change introduced in this PR is to clean up the console when the cwd is printed. Specifically when running `very_good test`.
Sometimes the full path of the cwd is too long and would be cut off. If running recursively, this can leave the developer confused as to which directory was tested.

image

A solution to the problem is to shorten the path printed to console. Instead of printing the absolute path, the relative path is used.

image

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/cli/flutter_cli.dart
+++ b/lib/src/cli/flutter_cli.dart
@@ -168,9 +168,8 @@ class Flutter {
         final target = DirectoryGeneratorTarget(Directory(p.normalize(cwd)));
         final workingDirectory = target.dir.absolute.path;
         final relativePath = p.relative(workingDirectory, from: initialCwd);
-        final relativePathPrefix = '.${p.context.separator}';
         final path =
-            relativePath == '.' ? '.' : '$relativePathPrefix$relativePath';
+            relativePath == '.' ? '.' : '.${p.context.separator}$relativePath';
 
         stdout?.call(
           'Running "flutter test" in $path ...\n',
--- a/lib/src/cli/flutter_cli.dart
+++ b/lib/src/cli/flutter_cli.dart
@@ -168,14 +168,17 @@ class Flutter {
         final target = DirectoryGeneratorTarget(Directory(p.normalize(cwd)));
         final workingDirectory = target.dir.absolute.path;
         final relativePath = p.relative(workingDirectory, from: initialCwd);
+        final relativePathPrefix = '.${p.context.separator}';
+        final path =
+            relativePath == '.' ? '.' : '$relativePathPrefix$relativePath';
 
         stdout?.call(
-          'Running "flutter test" in .${p.context.separator}$relativePath...\n',
+          'Running "flutter test" in $path ...\n',
         );
 
         if (!Directory(p.join(target.dir.absolute.path, 'test')).existsSync()) {
           stdout?.call(
-            'No test folder found in .${p.context.separator}$relativePath\n',
+            'No test folder found in $path\n',
           );
           return ExitCode.success.code;
         }
--- a/lib/src/cli/flutter_cli.dart
+++ b/lib/src/cli/flutter_cli.dart
@@ -153,6 +153,8 @@ class Flutter {
     FlutterTestRunner testRunner = flutterTest,
     GeneratorBuilder buildGenerator = MasonGenerator.fromBundle,
   }) async {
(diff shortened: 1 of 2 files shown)


lib/src/cli/flutter_cli.dart
```diff
@@ -165,14 +167,15 @@ class Flutter {
         void noop(String? _) {}
         final target = DirectoryGeneratorTarget(Directory(p.normalize(cwd)));
         final workingDirectory = target.dir.absolute.path;
+        final relativePath = p.relative(workingDirectory, from: initialCwd);
 
         stdout?.call(
-          'Running "flutter test" in ${p.dirname(workingDirectory)}...\n',
+          'Running "flutter test" in .${p.context.separator}$relativePath...\n',
```
alestiago, line 173: We should probably add a space between `$relativePath` and the `...` used as progress indicators to avoid confusing the user with four consecutive fullstops `....` when `$relativePath` is "." (as in some of the tests).

```suggestion
          'Running "flutter test" in .${p.context.separator}$relativePath ...\n',
```

alestiago, line 173: **Nit:**

We should include a space between the `$relativePath` and the `...` that indicate progress. To avoid confusion between the directory and the `...`.

```suggestion
          'Running "flutter test" in .${p.context.separator}$relativePath ...\n',
```

lib/src/cli/flutter_cli.dart
```diff
@@ -165,14 +167,15 @@ class Flutter {
         void noop(String? _) {}
         final target = DirectoryGeneratorTarget(Directory(p.normalize(cwd)));
         final workingDirectory = target.dir.absolute.path;
+        final relativePath = p.relative(workingDirectory, from: initialCwd);
```
alestiago, line 170: **Nit:**

Introduce a variable `relativePathPrefix` and use it to avoid duplicating the logic for the prefix on both calls.

```suggestion
        final relativePath = p.relative(workingDirectory, from: initialCwd);
        final relativePathPrefix = .${p.context.separator};
```

When doing so we should probably refactor the tests to include a method or variable, for example something as:
```dart
final relativePathPrefix = '.${p.context.separator}';
expect(
  stdoutLogs,
  equals([
    'Running "flutter test" in ${relativePathPrefix}foo ...\n',
```

test/src/cli/flutter_cli_test.dart
```diff
@@ -462,7 +462,7 @@ void main() {
         expect(
           stdoutLogs,
           equals([
-            'Running "flutter test" in ${p.dirname(tempDirectory.path)}...\n',
+            'Running "flutter test" in .${p.context.separator}....\n',
```
alestiago, line 465: I think reading `./.`  as path is not ideal. I think we should avoid prefixing `./` and simply report `.` in this scenario.

alestiago (commented):
Thank you for your contribution! I left some very small comments

test/src/cli/flutter_cli_test.dart
mrgnhnt96: Yeah, I can see that! `.` is a bit easier on the eyes than `./.`

lib/src/cli/flutter_cli.dart
mrgnhnt96: I like it! It would make it easier to read

mrgnhnt96:
Thanks for your review! See the respective changes and lmk if you have any other suggestions!

alestiago (approved):

lib/src/cli/flutter_cli.dart
```diff
@@ -165,14 +167,18 @@ class Flutter {
...
+        final relativePathPrefix = '.${p.context.separator}';
+        final path =
+            relativePath == '.' ? '.' : '$relativePathPrefix$relativePath';
```
alestiago, line 173: **Nit:** Just so we don't compute it before needing it. 
```suggestion
        final path =
            relativePath == '.' ? '.' : '.${p.context.separator}$relativePath';
```

wolfenrain (dismissed):

alestiago (approved):

Issue #901, opened 2023-12-01
fix: check licenses fails when project has "direct overridden" dependencies


mvolpato:
**Description**
When running `very_good packages check licenses` on a project whose `pubspec.lock` has  "direct overridden" dependencies, the command fails with "Null check operator used on a null value".

Tested with `very_good --version` 0.18.0.

**Steps To Reproduce**

1. Set a project with overridden dependencies;
2. Run `very_good packages check licenses`;

**Expected Behavior**
A report on licenses for the project.

**Actual Behaviour**

```bash
Null check operator used on a null value
#0      StringToDependencyType.parseDependencyType (package:pubspec_lock/src/package_dependency/dependency_type/serializers.dart:31:67)
#1      loadPathPackageDependency (package:pubspec_lock/src/package_dependency/path_package_dependency/serializers.dart:42:54)
#2      loadPackageDependency (package:pubspec_lock/src/package_dependency/serializers.dart:45:35)
#3      MappedIterator.moveNext (dart:_internal/iterable.dart:403:20)
#4      WhereIterator.moveNext (dart:_internal/iterable.dart:450:22)
#5      Iterable.length (dart:core/iterable.dart:524:15)
#6      PackagesCheckLicensesCommand.run (package:very_good_cli/src/commands/packages/commands/check/commands/licenses.dart:221:90)
<asynchronous suspension>
#7      CommandRunner.runCommand (package:args/command_runner.dart:212:13)
<asynchronous suspension>
#8      VeryGoodCommandRunner.runCommand (package:very_good_cli/src/command_runner.dart:126:18)
<asynchronous suspension>
#9      VeryGoodCommandRunner.run (package:very_good_cli/src/command_runner.dart:72:14)
<asynchronous suspension>
#10     main (file:///Users/michele/.pub-cache/hosted/pub.dev/very_good_cli-0.18.0/bin/very_good.dart:5:24)
<asynchronous suspension>
```

**Additional Context**
I removed all the "direct overridden" entries in my `pubspec.lock` and the command run successfully.

I think the issue might be fixed when a new version of `pubspec_lock` is released, see https://github.com/alexei-sintotski/pubspec_lock/issues/33. This issue has a PR that solves it.

alestiago:
Thanks @mvolpato for opening this issue and letting us know. 

[...]

PR #931, opened 2023-12-19
feat: support 'direct-overridden' dependency type when checking licenses

alestiago:
> Should be merged after #930 

## Status

**READY**

## Description

Adds support for "direct-overriden" dependency type on `very_good packages check licenses --dependency-type 'direct-overridden`.

Had to abandon the use of `pubspec_lock` since the dependency seems to not be maintained any more and overriding it caused the pana score to drop.

Somewhat related to #901 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/pubspec_lock/pubspec_lock.dart
+++ b/lib/src/pubspec_lock/pubspec_lock.dart
@@ -1,0 +1,186 @@
+/// A simple parser for pubspec.lock files.
+///
+/// This is used by the `packages check license` command to check the type and
+/// source of the dependencies to analyze. Hence it is not a complete parser,
+/// it only parses the information that is needed for the
+/// `packages check license` command.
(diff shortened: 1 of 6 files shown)


lib/src/pubspec_lock/pubspec_lock.dart
```diff
@@ -0,0 +1,186 @@
+/// A simple parser for pubspec.lock files.
+///
+/// This is used by the `packages check license` command to check the type and
+/// source of the dependencies to analyze. Hence it is not a complete parser,
```
alestiago, line 4:
```suggestion
/// source of the dependencies to analyze. Hence, it is not a complete parser,
```

wolfenrain (approved):

Issue #969, opened 2024-02-25
feat: add `.` as a valid option to `create` or allow no argument


a-wallen:
**Description**
As a user of the `very_good_cli` I want to be able to use `.` as the argument for `very_good create`. 

If no argument or `.` is given, the cli uses the current folder's name as the project name.

tomarra:
**Notes**

- The `flutter create` command works the same way, allowing `.` to automatically pull in the current directory name
- This should not override the current default which is to error out to the help output. If a user wants to use this option they should be explicitly adding `.` to the command.

tomarra:
Hi @a-wallen 👋 given the notes we have added here would you be willing to help out and submit a PR for this change?

a-wallen:
yeah I'm down

omartinma:
This issue got solved as part of the PR #996

PR #996, opened 2024-03-19
feat: allow `very_good create .`

a-wallen:
## Status

**READY/IN DEVELOPMENT/HOLD**
IN DEVELOPMENT

## Description

Allows `.` as a project name (like `flutter create .`)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/commands/create/commands/create_subcommand.dart
+++ b/lib/src/commands/create/commands/create_subcommand.dart
@@ -121,12 +121,23 @@ abstract class CreateSubCommand extends Command<int> {
     return Directory(directory);
   }
 
-  /// Gets the project name.
-  String get projectName {
+  /// The project name that the user provided.
+  ///
+  /// Since the project name could be user provided as either:
+  /// 1. A valid package name
+  /// 2. '.', the current working directory (the project assumes the cwd's name)
+  /// this needs to exist to provide a fallback value for [outputDirectory] and
+  /// [projectName].
+  String get _projectName {
     final args = argResults.rest;
     _validateProjectName(args);
     return args.first;
   }
+
+  /// Gets the project name.
+  String get projectName => _projectName == '.'
+      ? path.basename(Directory.current.path)
+      : _projectName;
 
   /// Gets the description for the project.
   String get projectDescription => argResults['description'] as String? ?? '';
@@ -159,6 +170,15 @@ abstract class CreateSubCommand extends Command<int> {
     }
 
     final name = args.first;
+
+    if (name == '.') {
+      logger.detail(
+        'Since the project name is ".", the current directory will'
+        ' be used as the project name.',
+      );
+      return;
+    }
+
     final isValidProjectName = _isValidPackageName(name);
     if (!isValidProjectName) {
       usageException(
@@ -211,7 +231,12 @@ abstract class CreateSubCommand extends Command<int> {
 
     await template.onGenerateComplete(
       logger,
-      Directory(path.join(target.dir.path, projectName)),
+      Directory(
+        [
+          target.dir.path,
+          if (_projectName != '.') projectName,
+        ].join(Platform.pathSeparator),
+      ),
     );
 
     return ExitCode.success.code;
--- a/test/src/commands/create/create_subcommand_test.dart
+++ b/test/src/commands/create/create_subcommand_test.dart
@@ -4,6 +4,7 @@ import 'package:args/args.dart';
 import 'package:args/command_runner.dart';
 import 'package:mason/mason.dart';
 import 'package:mocktail/mocktail.dart';
+import 'package:path/path.dart' as path;
 import 'package:test/test.dart';
 import 'package:very_good_cli/src/commands/create/commands/create_subcommand.dart';
 import 'package:very_good_cli/src/commands/create/templates/template.dart';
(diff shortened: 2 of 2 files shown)


test/src/commands/create/create_subcommand_test.dart
```diff
@@ -285,7 +286,61 @@ Run "runner help" to see global options.''';
                 that: isA<Directory>().having(
                   (d) => d.path,
                   'path',
-                  'test_dir/test_project',
+                  path.join('test_dir', 'test_project'),
```
a-wallen, line 289: There are probably other places where this exists `/`. 

Without `path.join` the tests are broken on Windows.

a-wallen:
fixes https://github.com/VeryGoodOpenSource/very_good_cli/issues/969

a-wallen:
this is ready for review.

by the way, the repo had 37 broken tests when I cloned it on my machine. I was wondering if I should submit those as issues.

tomarra:
Fixed the formatting issue by updating from `main` via #1003 so CI is now passing. @wolfenrain can we get a review here?

lib/src/commands/create/commands/create_subcommand.dart
```diff
@@ -121,13 +121,24 @@ abstract class CreateSubCommand extends Command<int> {
...
     final args = argResults.rest;
     _validateProjectName(args);
     return args.first;
   }
 
+  /// Gets the project name.
+  String get projectName => _projectName == '.'
+      ? path.basename(Directory.current.path)
+      : _projectName;
+
```
wolfenrain, line 141: I think we should rework this. Introducing `.` as a `projectName` and using it to find the name of the directory is kinda confusing because what if I pass `path/to/my/folder` instead?

`.` implies current directory which implicitly also means we can pass any path, which isnt the case. However this is the case for `flutter create` so maybe we should just make it that we can pass any given path and use the last part of that path as the `projectName` and create the template at the given path.

What you think?

a-wallen: Hi @wolfenrain, thanks for the positive feedback.

I agree, let's do it the same way as the Flutter tool.

This means we'd remove `--output-directory` as an argument. 

As far as `very_good create <arg>`, here are some requirements that I could implement:
- [ ] Allow relative paths as `arg`.
- [ ] Allow absolute paths as `arg`.
- [ ] Use the last path segment of `arg` as the project name.
- [ ] Remove the `--output-directory` argument.

Let me know if you'd like me to proceed. Thanks for catching my oversight 👍

wolfenrain: I do think that is a good approach, so `arg` would be any type of path indeed, it would be nice if we can keep backwards compatibility with the `--output-directory`. Maybe by just checking if it is supplied, if it is show that it is deprecated and just take that over whatever was in `arg`, assuming `arg` was a path?

PS: it would be `very_good create <template> <output directory>`

wolfenrain: Hey @a-wallen just following up here to see if you are planning on tackling this still?

lib/src/commands/create/commands/create_subcommand.dart
```diff
@@ -117,15 +117,43 @@ abstract class CreateSubCommand extends Command<int> {
...
-    return Directory(directory);
+    final directory = argResults['output-directory'] as String?;
+
+    if (directory != null) {
+      return Directory(directory);
+    }
+
+    final args = argResults.rest;
+
+    return Directory(args.first);
```
a-wallen, line 128: same logic as before, if the --output-directory argument isn't provided then treat the [project-name] argument as as a directory path.

test/src/commands/create/create_subcommand_test.dart
```diff
@@ -283,9 +411,13 @@ Run "runner help" to see global options.''';
               logger,
               any(
                 that: isA<Directory>().having(
-                  (d) => d.path,
+                  (d) {
+                    print(Directory.current);
+                    print(d.absolute.path);
```
a-wallen, line 416:
```suggestion
```

alestiago:
Pinging @wolfenrain

wolfenrain (approved):
LGTM, sorry for the late follow-up on my part. I will try and test out this branch later today to make sure it is backwards compatible but from the looks of it we should be good!

a-wallen:
sure thing, let me know if there's anything I need to do.

wolfenrain:
> sure thing, let me know if there's anything I need to do.

The CI seems to be failing on some tests, mind having a look at that? I suspect that the tests are failing because of wrong parameters (might mean it isnt backwards compatible like we hoped?)

a-wallen:
I can't tell which tests are failing. See this issue for more details https://github.com/VeryGoodOpenSource/very_good_cli/issues/1162

marcossevilla:
hey @a-wallen, can you take a look at the CI failures so we can merge this? thanks!

marcossevilla:
hey @a-wallen! do you need any help with this? I'd be happy to help you move past the finish line 👍

omartinma (approved):

PR #1084, opened 2024-07-12
fix: show a better message when coverage is really close to min coverage

Kirpal:
## Status

**READY**

## Description

On our project we had one line uncovered, which gave us a coverage of 99.995%. Our minimum coverage is 100%. The message given was `Expected coverage >= 100.00% but actual is 100.00%`, which isn't very helpful. This change will still use 2 decimal places unless the actual coverage is so close that it rounds to the expected amount, and in that case will increase to an appropriate number of decimal places.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/test/src/commands/test/test_test.dart
+++ b/test/src/commands/test/test_test.dart
@@ -413,6 +413,44 @@ void main() {
       ).called(1);
     });
 
+    test('displays required precision see why coverage was not met', () async {
+      when<dynamic>(() => argResults['coverage']).thenReturn(true);
+      when<dynamic>(() => argResults['min-coverage']).thenReturn('100');
+      const exception = MinCoverageNotMet(99.999995);
+      when(
+        () => flutterTest(
+          cwd: any(named: 'cwd'),
+          recursive: any(named: 'recursive'),
+          collectCoverage: any(named: 'collectCoverage'),
+          optimizePerformance: any(named: 'optimizePerformance'),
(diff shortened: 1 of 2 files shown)


test/src/commands/test/test_test.dart
```diff
@@ -413,6 +413,44 @@ void main() {
...
+          arguments: defaultArguments,
+          minCoverage: 100,
+          logger: logger,
+          stdout: logger.write,
+          stderr: logger.err,
+        ),
+      ).called(1);
+      verify(
+        () => logger.err(
+          'Expected coverage >= 100.000000% but actual is 99.999995%.',
```
wolfenrain, line 449: I would still prefer the 100 to stay the same (`100.00%`), the actual is where the decimal point matters.

Kirpal: I did this because there could be a case where someone has min coverage going to more than two decimal places, like `99.004`. In that case, if the coverage was 99.003, then the message would be `Expected coverage >= 99.00% but actual is 99.003%.`

Obviously it's an edge case, but I don't think we should round either value if it could make the error messaging more confusing. Plus, in most cases the coverage won't be that close and it will display 100.00% anyway.

alestiago (approved):

wolfenrain (approved):

PR #1333, opened 2025-08-08
feat: Adding the dart test command

erickzanardo:
## Status

**READY**

## Description

This PR adds a new `dart` command, which will center all dart only commands.

To start, it will include only the `dart test` command, which is the counterpart of the default test command, but runs `dart test` under the hood, instead of running `flutter test`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/cli/dart_cli.dart
```diff
@@ -92,4 +92,41 @@ class Dart {
 
     await Future.wait<void>(processes);
   }
+
+  /// Run tests (`flutter test`).
```
omartinma, line 96: Shouldnt be dart test?

test/src/command_runner_test.dart
```diff
@@ -34,6 +34,7 @@ const expectedUsage = [
   'Available commands:\n',
   '  create     very_good create <subcommand> <project-name> [arguments]\n',
   '''             Creates a new very good project in the specified directory.\n''',
+  '  dart       Command for running dart related commands.\n',
   '  packages   Command for managing packages.\n',
   '  test       Run tests in a Dart or Flutter project.\n',
```
matiasleyba, line 39: Should we update this description to clarify that there is a dart command that can be used for dart projects?

lib/src/cli/common.dart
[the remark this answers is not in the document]
erickzanardo: Oh, yes, forgot to remove it

PR #1444, opened 2025-12-29
feat: add workspace support for packages check licenses

realmeylisdev:
## Summary

This PR adds workspace support to the `packages check licenses` command, enabling license checking in monorepo projects that use Dart's pub workspace feature.

Closes #1273

### New Files
- `lib/src/pubspec/pubspec.dart` - Pubspec parser for detecting workspace configurations
- `test/src/pubspec/pubspec_test.dart` - 19 tests for the Pubspec parser

### Modified Files
- `lib/src/commands/packages/commands/check/commands/licenses.dart` - Added workspace detection and dependency collection
- `test/src/commands/packages/commands/check/commands/licenses_test.dart` - Added 3 workspace support tests

## How It Works

1. **Auto-detection**: When `pubspec.yaml` contains a `workspace` property, it's detected as a workspace root
2. **Dependency collection**: Dependencies are collected from all workspace members' `pubspec.yaml` files
3. **Glob pattern support**: Workspace paths like `packages/*` are supported (Dart 3.11+)
4. **Nested workspaces**: Recursively handles nested workspaces
5. **Backwards compatible**: Non-workspace projects work exactly as before

## Test plan

- [x] All existing license tests pass (42 tests)
- [x] New workspace support tests pass (3 tests)
- [x] New Pubspec parser tests pass (19 tests)
- [x] Static analysis passes with no issues

PR #1540, opened 2026-03-18
feat: add workspace support for packages check licenses

### LINKED ISSUE #1273: feat: Support `packages check licenses` with workspaces
### Description

When setting up a mono-repo with Pub workspaces your project will look this this:

- app/
  - `pubspec.yaml`
- packages/
  - packageA/
    - `pubspec.yaml`
  - packageB/
    - `pubspec.yaml`
- `pubspec.yaml`
- `pubspec.lock`

Meaning you can't run `very_good_cli packages check licenses` in the individual projects/packages (there is no `pubspec.lock` file anymore).

And you can't run `very_good_cli packages check licenses` in the workspace root, as it will look something like:

```yaml
name: _
publish_to: none
environment:
  sdk: ^3.6.0
workspace:
  - app
  - packages/packageA
  - packages/packageB

# No direct dependencies
# dependencies:

dev_dependencies:
  build_runner: ^2.4.14
  freezed: ^2.5.7
  json_serializable: ^6.9.0
```

And as the dependencies are not installed in the `pubspec.yaml` file next to the `pubspec.lock` file, the output will be something like:

> very_good packages check licenses
> No hosted dependencies found in /Users/steve/flutter_monorepo of type: direct-main.

### Requirements

- [ ] Dependencies are also checked in downstream workspaces  
  (e.g. when a pubspec.yaml file has a `workspaces` properties, go through the list of directories and read the dependencies there as if they were in the root file).
- [ ] Multiple nested workspaces also work  
  (if the root-pubspec.yaml links to the app-pubspec.yaml and the app-pubspec.yaml links to an internal-package-pubspec.yaml, also the dependencies of the internal-package-pubspec.yaml should be considered)
- [ ] All CI/CD checks are passing.
- [ ] There is no drop in the test coverage percentage.

### Additional Context

_No response_

realmeylisdev:
## Summary

This PR adds workspace support to the `packages check licenses` command, enabling license checking in monorepo projects that use Dart's pub workspace feature.

Reopens #1444 (branch was restored per request).
Closes #1273

### New Files
- `lib/src/pubspec/pubspec.dart` - Pubspec parser for detecting workspace configurations
- `test/src/pubspec/pubspec_test.dart` - 19 tests for the Pubspec parser

### Modified Files
- `lib/src/commands/packages/commands/check/commands/licenses.dart` - Added workspace detection and dependency collection
- `test/src/commands/packages/commands/check/commands/licenses_test.dart` - Added 3 workspace support tests

## How It Works

1. **Auto-detection**: When `pubspec.yaml` contains a `workspace` property, it's detected as a workspace root
2. **Dependency collection**: Dependencies are collected from all workspace members' `pubspec.yaml` files
3. **Glob pattern support**: Workspace paths like `packages/*` are supported (Dart 3.11+)
4. **Nested workspaces**: Recursively handles nested workspaces
5. **Backwards compatible**: Non-workspace projects work exactly as before

## Test plan

- [x] All existing license tests pass (42 tests)
- [x] New workspace support tests pass (3 tests)
- [x] New Pubspec parser tests pass (19 tests)
- [x] Static analysis passes with no issues


samitsv:
any update on this @marcossevilla . Would be great to have this merged soon.

marcossevilla:
hey @samitsv, we're currently trying the branch against internal repos to verify the changes, I'll leave an update once we finish testing

brianegan:
@marcossevilla Functional testing done on my end! PR worked well for a workspace monorepo. Sorry for the delay here -- deadlines.

lib/src/commands/packages/commands/check/commands/licenses.dart
```diff
@@ -35,6 +36,10 @@ Future<detector.Result> Function(String, double)? detectLicenseOverride;
 @visibleForTesting
 const pubspecLockBasename = 'pubspec.lock';
 
+/// The basename of the pubspec file.
+@visibleForTesting
+const pubspecBasename = 'pubspec.yaml';
```
marcossevilla: if we have a library for pubspecs, maybe it's more natural for these variables to part of it

lib/src/commands/packages/commands/check/commands/licenses.dart
```diff
@@ -531,6 +573,72 @@ extension on List<Object> {
...
+///
+/// Returns `null` if the file doesn't exist or cannot be parsed.
+Pubspec? _tryParsePubspec(File pubspecFile) {
+  if (!pubspecFile.existsSync()) return null;
+  try {
+    return Pubspec.fromFile(pubspecFile);
+  } on PubspecParseException catch (_) {
+    return null;
+  }
+}
```
marcossevilla: same here, maybe it's better if it's part of pubspec library

lib/src/pubspec/pubspec.dart
```diff
@@ -0,0 +1,206 @@
...
+      yaml = loadYaml(content) as YamlMap;
+      // loadYaml throws TypeError when it fails to cast content as a YamlMap.
+      // YamlException is thrown when the content is not valid YAML.
+      // We need to catch both to provide a meaningful exception.
+      // ignore: avoid_catching_errors
+    } on TypeError catch (_) {
+      throw const PubspecParseException('Failed to parse YAML content');
+    } on YamlException catch (_) {
+      throw const PubspecParseException('Failed to parse YAML content');
+    }
```
marcossevilla:
```suggestion
    } on Exception catch (_) {
      throw const PubspecParseException('Failed to parse YAML content');
    }
```
since it's the same exception being thrown

test/src/pubspec/pubspec_test.dart
```diff
@@ -0,0 +1,319 @@
...
+// ignore_for_file: prefer_const_constructors
+
+import 'dart:io';
+
+import 'package:path/path.dart' as path;
+import 'package:test/test.dart';
+import 'package:very_good_cli/src/pubspec/pubspec.dart';
+
+void main() {
+  group('$Pubspec', () {
```
marcossevilla:
```suggestion
  group(Pubspec, () {
```

test/src/pubspec/pubspec_test.dart
```diff
@@ -0,0 +1,319 @@
...
+        );
+        final members = pubspec.resolveWorkspaceMembers(tempDirectory);
+
+        expect(members.length, equals(1));
+        expect(path.basename(members.first.path), equals('member'));
+      });
+    });
+  });
+
+  group('$PubspecParseException', () {
```
marcossevilla:
```suggestion
  group(PubspecParseException, () {
```

test/src/pubspec/pubspec_test.dart
```diff
@@ -0,0 +1,319 @@
...
+      );
+    });
+
+    test('toString returns class name when no message', () {
+      final exception = PubspecParseException();
+      expect(exception.toString(), equals('PubspecParseException'));
+    });
+  });
+
+  group('$PubspecResolution', () {
```
marcossevilla:
```suggestion
  group(PubspecResolution, () {
```

marcossevilla (changes requested):

marcossevilla:
@realmeylisdev CI is failing, also found we could just use https://pub.dev/packages/pubspec_parse instead of creating a pubspec library

realmeylisdev:
## CI Failure Analysis

The `e2e (docs_site)` failure is **unrelated to this PR's changes**. The `e2e (flutter_plugin)` cancellation is a cascade from `fail-fast: true` in the matrix strategy.

### Root Cause

The docs_site e2e test generates a Docusaurus project and runs `npm run build`. The build fails because webpack's `ProgressPlugin` API dropped support for the `name` option:

```
ValidationError: Invalid options object. Progress Plugin has been initialized
using an options object that does not match the API schema.
 - options has an unknown property 'name'.
```

This is an upstream Docusaurus/webpack incompatibility in the `very_good_docs_site` template — the fix belongs in [very_good_templates](https://github.com/VeryGoodOpenSource/very_good_templates), not here.

### PR Code Verification

- All 73 unit tests pass (pubspec parser + licenses command including workspace support)
- `dart analyze --fatal-infos --fatal-warnings` — clean
- `dart format` — clean
- All other CI checks (build on ubuntu/windows, all other e2e tests, pana, spell-check, link-check) pass

marcossevilla:
hey @realmeylisdev, we need to use `package:pubspec_parse` still

marcossevilla:
hey @realmeylisdev, are you able to work on the changes? otherwise we could close the PR and the team will tackle this issue separately

realmeylisdev:
Hey @marcossevilla — swapped the custom parser for `package:pubspec_parse` per review in 8ca2652. The new `lib/src/pubspec/pubspec.dart` is down to ~85 lines and only keeps the workspace glob/filesystem helpers that `pubspec_parse` does not provide (`tryParsePubspec`, `PubspecWorkspace` extension with `isWorkspaceRoot`/`isWorkspaceMember`, and `resolveWorkspaceMembers`). Net diff: −238 lines. Coverage is preserved at 100% on both changed files (`pubspec.dart` 29/29, `licenses.dart` 240/240).

The two failing e2e checks (`docs_site`, `flutter_plugin`) are identical to the failures already present on `main` — see run https://github.com/VeryGoodOpenSource/very_good_cli/actions/runs/24140371198 (from #1567 merge, 4 days ago). Both are template-repo issues (Docusaurus/webpack `ProgressPlugin` incompatibility and mocktail `when`/`verify` inference warnings respectively), not related to this PR.

Ready for another look.

samitsv:
Can we please get this reviewed :) @marcossevilla

marcossevilla:
@realmeylisdev CI should be fixed, you can update your branch to get the latest changes

realmeylisdev:
> @realmeylisdev CI should be fixed, you can update your branch to get the latest changes

done.

marcossevilla:
@vgvbot perform a code review focused on verifying nits and code readability

erickzanardo:
@vgvbot perform a code review focused on verifying nits and code readability

lib/src/pubspec/pubspec.dart
```diff
@@ -0,0 +1,84 @@
...
+  bool get isWorkspaceRoot => workspace != null;
+
+  /// Whether this pubspec is a workspace member.
+  bool get isWorkspaceMember => resolution == _workspaceResolution;
+}
+
+/// Attempts to read and parse a [Pubspec] from [file].
+///
+/// Returns `null` when the file does not exist or cannot be parsed.
+Pubspec? tryParsePubspec(File file) {
```
marcossevilla: additional to the bot's comments, I'd say all the pubspec related methods can be in the existing PubspecWorkspace extension since we're just extending `pubspec_parse`

marcossevilla:
hey @realmeylisdev, will you be able to apply the comments on this PR? thanks!

realmeylisdev:
@marcossevilla applied the remaining bot review nits in d5668ab:

- Dropped `lenient: true` from `tryParsePubspec` so YAML-valid but structurally invalid pubspecs no longer silently parse with empty dep maps.
- Removed the unused `PubspecWorkspace.isWorkspaceMember` getter (dead in `lib/`, only tests referenced it).
- Replaced the hand-rolled `_isGlobPattern` heuristic with a single `Glob.listSync` path that handles literal and metachar patterns uniformly (so e.g. `!packages/foo` works without special-casing).
- Filled the remaining test gaps: structurally invalid pubspec, `?`/`[…]`/`{…}` glob shapes, literal paths with non-metachar symbols, and a member that omits `resolution: workspace`.

Ready for another look.

marcossevilla:
@vgvbot check the changes from your last review, resolve the comments that have been addressed and leave any comments if necessary

marcossevilla:
hey @realmeylisdev! let us know if you will be able to apply the changes, otherwise we can prioritize this change for the next CLI release (not #1605, but the following one)

realmeylisdev:
Thanks @marcossevilla — I've applied the changes (pushed in d2b106d). Summary of this round:

**`lib/src/pubspec/pubspec.dart`**
- Moved the parser into the extension (`PubspecWorkspace.tryParse`) so the pubspec helpers live together.
- `collectWorkspaceDependencies` now returns a non-null `Set` (empty when not a workspace); workspace detection reads `isWorkspaceRoot`, removing the previous null/Set dual signal.
- Workspace-root members are collected via the recursive call only (no more double-add), extracted into a `_directDependencies` helper.
- Documented the intentional `dependency_overrides` skip, the strict/silent parse behavior, and the POSIX glob assumption.

**`lib/src/commands/packages/commands/check/commands/licenses.dart`**
- Unified the dependency filter so the `transitive`/`direct-overridden` handling lives in one place; workspace-vs-direct is decided via `isWorkspaceRoot`.
- Reverted the `targetDirectory.path` → `targetPath` diff noise.

**Tests**
- The circular-workspace test now asserts on the collected set instead of only `returnsNormally`.
- Renamed the mixed-deps fixture and added an `app`-named fixture so member names match their directories.
- Confirmed coverage for the direct-main-in-one-member / direct-dev-in-another and YAML-valid-but-pubspec-invalid cases.

100% coverage is maintained, and `dart analyze --fatal-infos --fatal-warnings` and `dart format` are clean. Happy to adjust anything further.

realmeylisdev:
Heads up on CI: the failing `e2e (test/commands/create/flutter_plugin/flutter_plugin_test.dart)` check is inherited from `main`, not introduced by this PR.

- The same job is red on the #1620 merge commit on `main` ([run 27950506626](https://github.com/VeryGoodOpenSource/very_good_cli/actions/runs/27950506626)); every prior `main` commit passed it.
- The failure is in the generated flutter_plugin example (`check_platform_name`), where the test references `isMacOS`/`isAndroid`/`isWeb`/`isWindows` named parameters that aren't defined — template code this PR doesn't touch (it only changes `lib/src/pubspec/`, the licenses command, and their tests).
- The checks that exercise this PR's code all pass: `build` on ubuntu **and** windows, `pana`, `check_links`, `spell-check`, and the `licenses` e2e tests.

Re-merging `main` once that template issue is fixed should turn the check green.

realmeylisdev:
Thanks @marcossevilla and @unicoderbot for the thorough review. The feedback has been addressed in the latest revision — a summary of each point and how it was resolved:

### Architecture / API
- **Move workspace logic into the pubspec library** — all workspace traversal now lives in the `PubspecWorkspace` extension in `lib/src/pubspec/pubspec.dart` (`tryParse`, `isWorkspaceRoot`, `resolveMembers`, `collectWorkspaceDependencies`). The command file only calls into it.
- **Coupled signals via a nullable return** — `collectWorkspaceDependencies` now returns a `Set<String>` unconditionally (empty when not a workspace). The command reads workspace mode directly from `pubspec?.isWorkspaceRoot ?? false`, so there's no `!= null` overloading at the call site.
- **Redundant `direct-main`/`direct-dev` outer `if`** — the filter is now a single expression; in workspace mode membership is decided by `workspaceDependencies.contains(...)`, with `transitive`/`direct-overridden` handled in one place shared with the non-workspace path.

### Correctness / robustness
- **No cycle protection** — recursion now threads a `visited` set of canonicalized paths and short-circuits on revisit, so a member that points back at the root (or itself) terminates. Covered by a dedicated test that also asserts the cycling member's dep is still collected.
- **`lenient: true` masking malformed pubspecs** — dropped; parsing is now strict, so a structurally invalid pubspec is treated as unparseable rather than silently yielding empty dep maps. The dartdoc on `tryParse` states explicitly that parse failures are swallowed and malformed members are skipped.
- **Hand-rolled glob heuristic (`_isGlobPattern`)** — removed. Patterns go straight through `Glob(...).listSync(...)`, which handles literal vs. pattern paths uniformly (including negation), so the fragile `!`/`{`/`[` classification is gone.
- **Dead `isWorkspaceMember`** — removed (YAGNI).
- **Redundant add on nested workspace roots** — when a member is itself a workspace root, its direct deps are no longer added twice; the direct add is skipped and the recursive call handles them, with a comment explaining the flow.
- **`dependency_overrides` handling** — documented as intentional on `collectWorkspaceDependencies`: overrides are surfaced by the caller through the lock file's `direct-overridden` marker rather than from member pubspecs.
- **Workspace vs. lock-file "direct" divergence** — called out in a comment at the filter site so the behavior is discoverable.

### Tests
- **Cross-member `direct-main` / `direct-dev`** — added a test pinning the behavior: a dep declared `direct-main` in one member and `direct-dev` in another is still surfaced when only `--dependency-type direct-main` is requested.
- **Cycle test was vacuous (`returnsNormally`)** — now also asserts the returned set contains the cycling member's dependency.
- **Fixture naming** — `_workspaceMemberWithDevDeps…` renamed to `_workspaceMemberWithMixedDeps…`; fixtures whose `name:` didn't match their directory were fixed so member scenarios read naturally.
- **YAML-valid-but-pubspec-invalid case** — covered by the strict-parse change plus the malformed-member test.

### One intentional divergence — Windows globs
The suggestion to pin `Glob(pattern, context: p.posix)` was **not** applied, and I'd like to keep it that way: `listSync(root:)` walks the real filesystem, whose entity paths use the platform separator. Pinning POSIX would make pattern/path matching disagree with the filesystem on Windows. The platform-default context already accepts the forward-slash patterns that pub workspaces always use, on every platform — so the default is the more portable choice here. I've left a comment at that site explaining the reasoning.

CI is green and the full suite passes locally (analyzer clean with `--fatal-infos --fatal-warnings`; workspace-support and pubspec tests all pass). Re-requesting a review when you have a chance 🙏

realmeylisdev:
Rebased onto the latest `main` — the branch is now up to date with no conflicts, and CI is green across all checks. All review threads from the earlier round are resolved, and @unicoderbot approved the latest revision.

@marcossevilla, when you have a moment, could you take another look? Happy to address anything further. Thanks!

marcossevilla:
hi @realmeylisdev, thanks for your contribution but we will move forward with the implementation on #1657 since it's simpler and more inline with our practices and standards.

thanks again for your initiative and we will welcome other PRs in the future!

PR #1595, opened 2026-05-13
feat(test): add timeout

eliasreis54:
## Status

**READY**

## Description

Flutter tests can hang indefinitely when pumpAndSettle() is called without a timeout argument. The MCP server was waiting on the process forever with no way to recover.

Adds a `timeout_seconds` parameter (default 120) to the MCP test tool and a `--timeout` flag to `very_good test`. The value is passed directly to `flutter test` as `--timeout=<seconds>s`, delegating timeout enforcement to package:test.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


ryzizub (commented):
Hey @eliasreis54, this might fix the MCP issue where timeouts from `flutter test` aren't bubbled up. But since the same issue exists in the `very_good test` command itself, shouldn't the fix be applied there too?

There we might find better way to tackle this, because this might introduce many issues (120 seconds for large codebases is quite aggresive)

ryzizub (approved):
LGTM, only update of description on PR needed

marcossevilla (approved):

PR #1600, opened 2026-05-29
fix(mcp): apply `directory` argument as the working directory

### LINKED ISSUE #1599: vgv-ai-flutter-plugin feat #70 doesn't work
### Description

The `directory` argument on the `test`, `packages_get`, and `packages_check_licenses`
MCP tools is appended as a **positional** CLI argument. Those commands resolve their
target package from `Directory.current` and treat positionals as test/target paths, so
the value never changed *where* the command ran.

In a monorepo, the tools therefore execute from the MCP server's current directory
(commonly a folder with no `pubspec.yaml`) and fail with exit code 66 — regardless of
whether a relative or absolute `directory` was supplied. This defeats the purpose of the
`directory`/`path` parameter introduced for monorepo support (https://github.com/VeryGoodOpenSource/vgv-ai-flutter-plugin/issues/70).

### Steps To Reproduce

1. Have an agent run a very_good_cli MCP 'test' tool on a subfolder of a monorepo project which has no root `pubspec.yaml`.
2. Get an exit code 66.

### Expected Behavior

Use the `directory` argument as the working directory, not a parameter to pass on to the underlying tool.

### Additional Context

I and Claude have already patched my local copy and it now works as expected, so I can easily create a PR from that.

jmadren:
## Description

The `directory` argument on the `test`, `packages_get`, and `packages_check_licenses` MCP tools was appended as a positional CLI argument. Those commands resolve their target package from `Directory.current` and treat positionals as test/target paths, so the value never changed where the command ran. In a monorepo the tools therefore executed from the MCP server's current directory (commonly a folder with no `pubspec.yaml`) and failed with exit code 66, regardless of whether a relative or absolute `directory` was supplied.

Closes #1599

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

## Fix

Apply `directory` by switching `Directory.current` for the duration of the run and restoring it in a `finally` afterwards. Relative paths resolve against the server's current directory, matching the intent of the `path`/`directory` parameter from https://github.com/VeryGoodOpenSource/vgv-ai-flutter-plugin/issues/70. A non-existent directory now surfaces a clear tool error instead of a confusing exit code.

Refs https://github.com/VeryGoodOpenSource/vgv-ai-flutter-plugin/issues/70

### DIFF
--- a/lib/src/mcp/mcp_server.dart
+++ b/lib/src/mcp/mcp_server.dart
@@ -1,5 +1,5 @@
 import 'dart:async';
-import 'dart:io' show stderr;
+import 'dart:io' show Directory, stderr;
 
 import 'package:args/command_runner.dart';
 import 'package:dart_mcp/server.dart';
@@ -322,11 +322,10 @@ Only one value can be selected.
   }
 
   List<String> _parseTest(Map<String, Object?> args) {
+    // NOTE: 'directory' is intentionally not added here. It is applied as the
+    // working directory in [_runToolCommand], not as a positional test target.
     final cliArgs = <String>[if (args['dart'] == true) 'dart', 'test'];
 
-    if (args['directory'] != null) {
-      cliArgs.add(args['directory']! as String);
-    }
     if (args['coverage'] == true) {
       cliArgs.add('--coverage');
     }
@@ -389,11 +388,10 @@ Only one value can be selected.
   }
 
   List<String> _parsePackagesGet(Map<String, Object?> args) {
+    // NOTE: 'directory' is applied as the working directory in
+    // [_runToolCommand], not as a positional argument.
     final cliArgs = <String>['packages', 'get'];
 
-    if (args['directory'] != null) {
-      cliArgs.add(args['directory']! as String);
-    }
     if (args['recursive'] == true) {
       cliArgs.add('--recursive');
     }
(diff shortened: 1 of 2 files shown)


ryzizub:
Thank you @jmadren for the fix and contribution! Let me take a look and test and i will get back to you 💙

ryzizub (commented):
Hey, this is an awesome contribution and it actually solves the issue!
But I think we could add that option to the test command itself and not doing this override. That way both the MCP and the command can change the directory and also select the test (which is what's working right now).
Would you be open to making that update?

jmadren:
I'm not sure I follow. I thought the issue was the MCP command didn't use the directory parameter correctly, so how would adding (another) option to the test command fix this?

marcossevilla (commented):
I'm good to approve/merge this, thoughts @ryzizub ?

ryzizub:
Yeah @marcossevilla, I'm good too. I'll do a couple of tests, but I think it's great :)

@jmadren just for context, the questions earlier were about the idea of adding a --directory option to the test command. But we agreed internally that your PR is good and less invasive ❤️

marcossevilla (approved):

ryzizub (approved):
LGTM, good job @jmadren

PR #1612, opened 2026-06-15
fix(mcp): surface command output and run tools in the requested directory

### LINKED ISSUE #1611: fix: MCP tools report only an exit code and discard the command output
### Description

The `test`, `packages_get`, and `packages_check_licenses` MCP tools run the CLI **in-process** and surface only a synthesized line such as `"test" failed with exit code 69.` in the `CallToolResult`. The actual command output — failing test names, assertion messages, compile errors, coverage details — is written by the in-process mason `Logger` to the MCP server's `stdout`/`stderr`, which under the stdio transport is shared with the JSON-RPC message stream. Non-MCP lines on `stdout` violate the stdio transport contract (the server MUST NOT write anything to `stdout` that is not a valid MCP message) and are dropped by the client, so the agent only ever sees the opaque exit code and cannot diagnose the failure.

Exit code `69` (`ExitCode.unavailable`) is especially unhelpful: the `test` command collapses every failure mode — compile error, failed assertion, unmet `--min-coverage`, thrown exception — into that single code (`lib/src/commands/test/test.dart`, `lib/src/cli/test_cli_runner.dart`).

### Steps To Reproduce

1. Run the very_good_cli MCP `test` tool on a package whose tests fail (or fail to compile).
2. Observe the tool result: `"test" failed with exit code 69.` with no failing test names, error messages, or compile output.

### Expected Behavior

The tool result should include the captured command output (the real failure detail) in the `CallToolResult` `content`, with `isError: true`, so the agent can act on it. Per the MCP spec, tool-execution failures are reported inside the result with `isError: true` and the human-readable detail in `content` — not as a bare status line.

### Additional Context

- Lives in the same handler (`_runToolCommand` in `lib/src/mcp/mcp_server.dart`) as the `directory` bug (#1599 / #1600).
- Related: VeryGoodOpenSource/vgv-ai-flutter-plugin#94 — agents get stuck because the MCP `test` result is opaque.
- Possible approaches:
  - Capture the in-process `Logger` output via an `IOOverrides` `stdout`/`stderr` redirect and include a bounded tail in `content`. Note: the `Logger` must be constructed **inside** the zone, since mason captures `IOOverrides.current` at construction time.
  - Or run the CLI as a subprocess with piped `stdout`/`stderr` (cleaner separation from the JSON-RPC stream, at the cost of in-process speed/testability).
  - Output should be size-bounded and scrubbed of secrets (e.g. `--dart-define` values, tokens in stack traces).
  - Optionally stream progress via the MCP `logging` capability (`notifications/message`) rather than the process `stderr`.

ryzizub:
## Status

**READY**

## Description

The very_good_cli MCP `test`, `packages_get`, and `packages_check_licenses` tools had two problems:

1. **Command output was discarded.** A failed run returned only `"<tool>" failed with exit code 69.` — the actual failing tests, compile errors, and logs (written by the in-process mason `Logger` to stdout/stderr, which the stdio transport shares with the JSON-RPC stream) never reached the client. Agents saw an opaque exit code and could not diagnose failures.
2. **`directory` did nothing useful.** It was appended as a positional CLI argument; the commands resolve their target from the process current directory, so in a monorepo the tool ran from the wrong place and failed with exit code 66.

### Changes

- Capture the in-process command output by redirecting stdout/stderr to a buffer via `IOOverrides.runZoned`, constructing the mason `Logger` **inside** the zone (mason pins `IOOverrides.current` at construction, so a Logger built outside would still write to the real stdout). The captured output is returned in `CallToolResult.content` with `isError` on success and every failure path, and non-JSON is kept off the real stdout that carries the MCP JSON-RPC stream (per the stdio transport spec).
- Apply `directory` as the real working directory (`Directory.current`) for the run and restore it in `finally`; remove the positional misuse.
- Serialize tool runs behind an async mutex so the process-global working-directory switch is safe under concurrent/pipelined tool calls (the json_rpc_2 transport can dispatch handlers concurrently).
- Ignore generated `*.vm.json` test artifacts.

### Connections

- Closes #1611
- Overlaps with #1600 / #1599: this PR also reworks `directory` into a working-directory switch. The approach here is concurrency-safe (serialized) and additionally surfaces command output — coordinate to avoid duplicate/conflicting changes.
- Refs VeryGoodOpenSource/vgv-ai-flutter-plugin#94 (downstream symptom: opaque MCP `test` results blocked the `/build` flow).

### Verification

- MCP suite: 40 tests pass, `lib/src/mcp/mcp_server.dart` at 100% coverage.
- `dart analyze --fatal-infos --fatal-warnings` clean; `dart format` clean.
- Verified live through the MCP `test` tool: a failing package now returns the real failure output and runs in the requested `directory`. The concurrency fix is covered by a mutation-proven regression test.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/src/mcp/mcp_server.dart
```diff
@@ -476,71 +506,209 @@ Only one value can be selected.
...
+
+        stderr.writeln(
+          '[very_good_mcp] "$toolName" failed with exit code '
+          '$exitCode. Command: $commandString',
+        );
+        return CallToolResult(
+          content: [
+            TextContent(
+              text: withCapturedOutput(
+                '"$toolName" failed with exit code $exitCode.\n'
```
omartinma: nit: we basically have the same error message here and in stderr.writeln, should we extract that to a variable?

omartinma (dismissed):

lib/src/mcp/mcp_server.dart
ryzizub: Good call. Extracted an `errorResult(reason, {stackTrace})` helper that builds the message once, logs it to stderr, and surfaces it (with captured output) in the result. The three catch branches now just pass the reason. Pushed.

marcossevilla (approved):

omartinma (approved):

PR #1676, opened 2026-08-05
feat(mcp): allow the `test` tool to target specific paths

### LINKED ISSUE #1675: feat: allow the `test` MCP tool to target specific paths
## Description

The `test` MCP tool has no argument for running a subset of a test suite, so every invocation runs the whole thing. There is no workaround through the existing arguments: `directory` is explicitly *not* forwarded as a positional target, per the note in `_parseTest`:

```dart
// NOTE: 'directory' is intentionally not added here. It is applied as the
// working directory in [_runToolCommand], not as a positional test target.
```

and `tags` / `exclude_tags` only filter by annotations that already exist in the test source.

The CLI itself has no such gap. `very_good test test/foo_test.dart` works today: `FlutterTestOptions.parse` keeps `argResults.rest` and the command forwards it to the runner. So this is a hole in the MCP surface only, and an agent driving the CLI through MCP is strictly less capable than one shelling out.

That matters for agent workflows in two ways:

1. **Latency.** On a mid-sized app suite, running everything to check one directory turns a ~10s feedback loop into a ~50s one.
2. **Context cost.** The tool returns the runner's full output. On a suite with a failure that produces a large widget-tree or stack dump, a single call can return well over 100k characters, nearly all of it irrelevant to the tests the caller cared about. Being able to narrow the run is the cheapest available mitigation.

## Requirements

- [ ] The `test` tool accepts optional test paths and forwards them as positional targets.
- [ ] Paths are appended after all options, so they are parsed as `rest` rather than consumed as an option value.
- [ ] Works for both `very_good test` and `very_good dart test`.
- [ ] Existing behaviour is unchanged when the argument is omitted.
- [ ] All CI/CD checks are passing.
- [ ] There is no drop in the test coverage percentage.

## Additional Context

One behaviour worth documenting in the argument description rather than working around: targeting specific files already disables the test optimization step, via `TestCLIRunner.isTargettingTestFiles(options.rest)` in both `test.dart` and `dart_test_command.dart`. That is existing CLI behaviour and should stay consistent — callers just need to know it.

I have a patch ready and will open a PR referencing this issue. Happy to adjust the argument name or shape (array vs. comma-separated string) to whatever you prefer.

### LINKED ISSUE #1704: fix: `very_good test --recursive` applies positional test targets to every package
## Description

Positional test targets are forwarded verbatim into every per-package run when `--recursive` is set, so a relative path can only resolve in one package and fails to load in all the others.

`_runCommand` in `lib/src/cli/flutter_cli.dart` fans out one run per discovered `pubspec.yaml`, each with that package as its working directory:

```dart
final processes = _Cmd.runWhere<T>(
  run: (entity) => cmd(entity.parent.path),
  where: (entity) => !ignore.excludes(entity) && _isPubspec(entity),
  cwd: cwd,
);
```

`TestCLIRunner.test` then passes the same `arguments` (which carry `argResults.rest`) into each of those runs unchanged.

## Steps To Reproduce

1. Create a monorepo with two packages, each with its own test:

```
my_monorepo/
  packages/
    a/pubspec.yaml   test/a_test.dart
    b/pubspec.yaml   test/b_test.dart
```

2. From `my_monorepo`, run:

```sh
very_good dart test -r test/a_test.dart
```

3. `packages/a` passes. `packages/b` tries to load `packages/b/test/a_test.dart`, which does not exist, fails to load, and the command exits 69.

## Expected Behavior

Either of the following would be reasonable:

1. **Reject the combination.** `--recursive` and positional test targets describe mutually exclusive models of the working directory: a target is only meaningful relative to a single package root, and `--recursive` means there is not one. Failing fast with a message pointing at the right invocation is more useful than N-1 load errors around one incidental success.

2. **Resolve targets against the invocation directory.** Interpret each target relative to the original cwd, then pass each package only the targets that fall inside it, skipping packages with no matches. This would make `very_good test -r packages/a/test/x_test.dart packages/b/test/y_test.dart` do the obvious thing, but it needs decisions about targets that match no package or sit outside all of them.

Option 1 is the smaller change. `TestCLIRunner.test` already receives both `recursive` and the arguments, and `TestCLIRunner.isTargettingTestFiles` already exists as the predicate, so the check has a natural home at the point both `very_good test` and `very_good dart test` route through.

## Additional Context

This came out of review on #1676, which adds a `paths` argument to the MCP `test` tool and so exposes both knobs side by side in a single call. That PR currently guards the combination in the MCP handler, but the invariant belongs to the test command rather than the transport — a guard here would cover both entry points and shell users.

Happy to open a PR once there is agreement on which direction you prefer.

HadiHassan22:
## Description

Closes #1675.

The `test` MCP tool had no way to run a subset of a suite, so every call ran the whole thing. There was no workaround through the existing arguments: `directory` is deliberately applied as the working directory rather than as a target (per the existing `NOTE` in `_parseTest`), and `tags` / `exclude_tags` only filter on annotations already present in the test source.

The CLI has no such gap — `very_good test test/foo_test.dart` works today, because the test commands keep `argResults.rest` and forward it to the runner. This just exposes that through MCP.

## Changes

- Adds an optional `paths` argument to the `test` tool: a list of strings, each a test file or directory.
- `_parseTest` appends them **after** every option, so they are parsed as `rest` rather than consumed as the value of a preceding option.
- Works for both `very_good test` and `very_good dart test`; omitting the argument leaves behaviour exactly as before.

The argument description also notes that targeting specific paths disables the test optimization step. That is pre-existing CLI behaviour via `TestCLIRunner.isTargettingTestFiles`, not something this PR changes — it just makes it discoverable to a caller who can no longer see the command line.

## Why this matters for MCP callers

An agent driving the CLI through MCP was strictly less capable than one shelling out. Beyond the wasted wall-clock of running everything to check one directory, the tool returns the runner's full output, so a suite whose failure produces a large widget-tree or stack dump can return well over 100k characters — nearly all of it irrelevant to the tests the caller cared about. Narrowing the run is the cheapest mitigation available.

## Testing

Three tests added to `test/src/mcp/mcp_server_test.dart`, covering both branches of the new code:

- paths forwarded as positional targets
- paths appended after options (asserts ordering with `dart` + `concurrency`)
- an empty list adds no positional targets

Verified locally:

- `dart format lib test` — no changes
- `dart analyze --fatal-infos --fatal-warnings lib test` — no issues
- `very_good dart test -x pull-request-only` — 516 passing

## Open questions

Happy to change the argument name (`paths` vs `targets` vs `test_paths`) or its shape — I used a string array, though `packages_get`'s `ignore` uses a comma-separated string, so let me know if you'd rather stay consistent with that.


jmadren:
### Some points to raise on this PR:

1. **`paths` is a footgun that reproduces the [#1600](https://github.com/VeryGoodOpenSource/very_good_cli/pull/1600) symptom.** From a monorepo root with no `pubspec.yaml`, `very_good dart test packages/a/test/a_test.dart` exits **66** — the exact failure from [#1599](https://github.com/VeryGoodOpenSource/very_good_cli/issues/1599)/[#1600](https://github.com/VeryGoodOpenSource/very_good_cli/pull/1600). An agent that puts a package path into `paths` instead of `directory` lands right back there. The description should say the paths are test targets *inside the package selected by `directory`*, and that a package path belongs in `directory`.

2. **`recursive: true` + `paths` is incoherent.** `TestCLIRunner` spawns per package with `workingDirectory: cwd` and passes `options.rest` verbatim, so the same relative path is applied in *every* package. Measured: `very_good dart test -r test/a_test.dart` passes in `packages/a`, fails to load in `packages/b`, exit **69**. Pre-existing CLI behavior, but the MCP surface now exposes both knobs side by side with no shell to make it obvious — and it contradicts the new arg's "relative to the project root" wording, which is only true when `recursive` is off.

3. **Minor:** emitting `['--', ...paths]` would be strictly more robust (a path beginning with `-` currently becomes a `UsageException`), and matches what `isTargettingTestFiles` documents. The `--` is stripped from `rest` by the parser, so nothing downstream changes.

4. **Minor:** `doc/mcp.md`'s `test` example JSON enumerates every argument and wasn't updated.

5. One inaccuracy in the *existing* codebase that the PR inherits rather than introduces: `TestCLIRunner.isTargettingTestFiles`'s doc comment says "the user can only target test files by using the `--` option terminator." That's false — `allowTrailingOptions` is on, so bare positionals work, which is exactly why the PR works at all.

ryzizub:
@HadiHassan22 thanks for the contribution, its a good catch! @jmadren left really nice notes can you check those? (thank you for the review A LOT)

HadiHassan22:
@jmadren Thank you for taking the time to review my PR.

Here's how I addressed your points:

1. Reworded `paths` — targets are now described as relative to the package selected by `directory`, with an explicit note that a package path belongs in `directory`. Also changed `directory` from the generic "Target directory path" to "The package to test", so the two read as a pair.
2. `recursive` + `paths` is now rejected in the CLI. More below.
3. Emits `['--', ...paths]`, guarded on non-empty so an empty list still gives bare `['test']`.
4. Added `paths` to the `doc/mcp.md` example, plus a short note on the `directory`/`paths` split. I flipped that example's `recursive` to `false` since it enumerates every argument and leaving it `true` would have documented the exact combination we now reject.
5. Fixed the `isTargettingTestFiles` comment in its own commit, since it's CLI-layer rather than MCP.

On point 2:

Passing the pair through applies the same relative path in every package, so it resolves in one and fails to load in the rest leading to N-1 spurious failures around one success. The combination also has no coherent meaning: `paths` is documented as relative to the package `directory` selects, and `recursive` means there isn't one.

My first pass guarded it in `_handleTest`, but the invariant belongs to the test command rather than the transport. Left there, the rule exists in one entry point and not the other, and it would silently block the combination if the CLI ever learned to handle it properly. So `very_good test` and `very_good dart test` now exit with a usage code and an explanation, the MCP handler goes back to building argv and nothing else, and both entry points inherit the same behaviour. Written up in #1704 .

One placement note, since it isn't the obvious spot: the check can't live in `TestCLIRunner.test`. Targets arrive there already merged into `arguments` alongside option values, and those values don't start with `-` (`['-x', 'e2e']`, `['-j', '4']`), so `isTargettingTestFiles` would match unconditionally. It sits in each command beside the existing `pubspec.yaml` guard, where `argResults.rest` is unambiguous.

The schema still documents the exclusion. That's what an agent reads before calling, rather than after a run fails.

Happy to trim this back to a doc note if you'd rather keep the PR purely MCP, in which case the CLI fix moves to #1704 on its own.

marcossevilla:
@HadiHassan22 we just merged the migration to very_good_analysis 11 changes, which might've introduced new lints or formatting changes, please let me know if you can update this PR with those changes so we can merge it ASAP! thanks

HadiHassan22:
@marcossevilla done, everything should be green with the new formatting and lints

ryzizub (approved):
LGTM

PR #1689, opened 2026-08-13
fix: resolve dart coverage package_config in pub workspaces

### LINKED ISSUE #1683: fix: very_good dart test --coverage fails in pub workspaces
### Description

Running `very_good dart test` with coverage collection enabled (via `--coverage`, or implicitly via `--min-coverage`/`--show-uncovered`, or via the new 1.4.0 `very_good.yaml` `dart.test.min_coverage` config) crashes with a `PathNotFoundException` when run from inside a member package of a [Dart pub workspace](https://dart.dev/tools/pub/workspaces).

Pub workspaces resolve dependencies once at the workspace root, so `.dart_tool/package_config.json` is only generated at the root, not duplicated inside each member package's own `.dart_tool/`. The `very_good dart test` coverage step looks for that file relative to the current working directory instead of resolving it the way `dart pub`/`dart test` themselves do (walking up to the workspace root), so it fails whenever it's invoked from inside a member directory.

### Steps To Reproduce

1. Create a pub workspace:
   ```yaml
   # pubspec.yaml (root)
   name: workspace_root
   environment:
     sdk: ^3.9.0
   workspace:
     - packages/foo
   ```
2. Create a member package `packages/foo` with `resolution: workspace` in its `pubspec.yaml`, and a trivial passing test.
3. From the workspace root, run `dart pub get`.
4. `cd packages/foo`
5. Run `very_good dart test --coverage`
6. See error:
   ```
   Running "dart test" in . ...
   ✓ Optimizing tests (90ms)
   00:01 +60: All tests passed!
   PathNotFoundException: Cannot open file, path = '.dart_tool/package_config.json' (OS Error: No such file or directory, errno = 2)
   ```
   Exit code 69.

### Expected Behavior

Tests run and a coverage report is produced, the same as running `dart test --coverage` directly (which works fine in this exact layout).

### Additional Context

- Setting `dart.test.min_coverage` in `very_good.yaml` implicitly turns on coverage collection (`collectCoverage: options.collectCoverage || options.minCoverage != null || options.showUncovered`), so a bare `very_good dart test` with no flags at all starts crashing the moment that config key is added, with no indication that pub workspaces are the cause.
- `very_good test` (the Flutter-oriented command, which delegates to `flutter test`) does not hit this issue in the same workspace/directory — its coverage-collection path resolves the package config differently and more tolerantly of the workspace layout.
- Confirmed the workspace only materializes `package_config.json` at the root (`workspace_root/.dart_tool/package_config.json`), never inside `packages/foo/.dart_tool/`, even after re-running `dart pub get` from inside the member directory.

**Possible fix**

`lib/src/cli/test_cli_runner.dart`, inside `TestCLIRunner.test()`, in the block that post-processes coverage for `dart test`:

```dart
if (testType == TestRunType.dart && collectCoverage) {
  final files = _dartCoverageFilesToProcess(p.join(cwd, 'coverage'));

  final packagesPath = p.join('.dart_tool', 'package_config.json');
  final hitmap = await coverage.HitMap.parseFiles(
    files,
    packagePath: packagesPath,
    checkIgnoredLines: checkIgnore,
  );

  final resolver = await coverage.Resolver.create(
    packagesPath: packagesPath,
    packagePath: packagesPath,
  );
  ...
```

Every other path in this function joins against `cwd` (e.g. `p.join(cwd, 'coverage', 'lcov.info')` a few lines above). This one is a bare literal resolved relative to whatever directory the process happens to be in, which never resolves in a pub workspace since `.dart_tool/package_config.json` only exists at the workspace root.

Two ways to fix it, in order of correctness:

1. Join against `cwd`: `p.join(cwd, '.dart_tool', 'package_config.json')`. Fixes non-workspace recursive runs against nested directories, but still doesn't fix workspaces.
2. Walk up from `cwd` looking for `.dart_tool/package_config.json`, the same way `VeryGoodConfig.load()` already walks up looking for `very_good.yaml` in `lib/src/very_good_config/very_good_config.dart`. That walk-up logic could be extracted and reused here.

**Environment**

- very_good_cli: 1.4.0
- Dart SDK: 3.12.2 (stable)
- Flutter: 3.44.2
- OS: macOS 15 (Darwin 25.5.0), arm64

AzazelSensei:
## Status

**READY**

## Description

`very_good dart test --coverage` crashes in a pub workspace because it opens `.dart_tool/package_config.json` relative to the process cwd. Workspaces only write that file at the root.

I pass the package directory as `packagePath` so the coverage package uses `findPackageConfig` and walks up, same as dart.

Fixes #1683

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


marcossevilla:
Nice fix, this is a cleaner approach than a hand-rolled walk-up. One test-coverage gap worth closing before merge:

The new test (`resolves dart coverage package_config from the pub workspace root`) proves the pipeline no longer crashes, but it never proves a `package:` URI actually resolves through the workspace-root config back to a real source file. The seeded `package_config.json` has an empty `"packages":[]`, and the `coverage/` dir never gets a real coverage JSON (only an empty `lcov.info`), so `HitMap.parseFiles` runs with zero entries either way. That's enough to catch the `PathNotFoundException` regression, but not enough to prove resolution actually works.

To close the gap, the test would need:

1. A real package entry in the workspace-root `package_config.json`, e.g.
   ```json
   {"configVersion":2,"packages":[{"name":"foo","rootUri":"../packages/foo","packageUri":"lib/"}]}
   ```
2. A real source file, e.g. `member/lib/foo.dart`.
3. A `coverage/*.json` file in the shape `HitMap.parseFiles` actually reads (`{"coverage":[...]}` per `hitmap.dart`):
   ```json
   {"coverage":[{"source":"package:foo/foo.dart","hits":[1,1]}]}
   ```

Then assert on the resulting `lcov.info` content (e.g. `contains('SF:lib/foo.dart')`) instead of just the exit code, to confirm the resolver actually mapped the `package:` URI back to the file via the workspace-root config.

Not blocking since none of the other dart-coverage tests in this file do this either (they all use empty/placeholder coverage data), but it's the one thing that would make this test prove the fix rather than just prove no crash.

marcossevilla:
hey @AzazelSensei, thanks for opening a PR! Will you be able to apply the change I mentioned above?

AzazelSensei:
Applied the coverage test change. It now seeds a real package entry, source file, and coverage JSON, then checks lcov.info for SF:lib/foo.dart. Rebased onto main.

marcossevilla:
@AzazelSensei fix is great! but tests are failing, let me know when they're passing so we can approve

AzazelSensei:
Windows was writing an empty lcov. reportOn now uses the resolved package path so the lib prefix still matches.

AzazelSensei:
Formatted the two files dart format was rewriting. The Windows empty lcov is the reportOn change already on the branch; needs another workflow approval to run.

marcossevilla:
@AzazelSensei currently the flame_game e2e is failing for all PRs, so just take care of the very_good_cli build workflow

AzazelSensei:
Build was failing analyze on prefer_if_elements_to_conditional_expressions in test_cli_runner. Switched that reportOn map to an if-element.

lib/src/cli/test_cli_runner.dart
```diff
@@ -198,25 +198,38 @@ class TestCLIRunner {
...
-                  );
+                  // Resolve package_config.json the way dart does: start at
+                  // the package cwd and walk up. In a pub workspace the file
+                  // lives at the workspace root, not in the member package.
+                  //
+                  // Canonicalize cwd first. On Windows the temp/workspace
+                  // directory is often a junction, and package:coverage
+                  // resolves files through that real path. reportOn: ['lib']
+                  // is canonicalized against the unresolved cwd, so the
+                  // prefix check drops every file and lcov comes out empty.
```
marcossevilla:
```suggestion
```

test/src/cli/test_cli_runner_test.dart
```diff
@@ -778,6 +778,73 @@ void main() {
...
+          final member = Directory(
+            p.join(workspaceRoot.path, 'packages', 'foo'),
+          )..createSync(recursive: true);
+          File(p.join(member.path, 'pubspec.yaml')).createSync();
+          Directory(p.join(member.path, 'test')).createSync();
+          File(p.join(member.path, 'lib', 'foo.dart'))
+            ..createSync(recursive: true)
+            ..writeAsStringSync('void foo() {}');
+
+          // Pub workspaces only write package_config.json at the root.
```
marcossevilla:
```suggestion
```

marcossevilla:
@AzazelSensei we just merged the migration to very_good_analysis 11 changes, which might've introduced new lints or formatting changes, please let me know if you can update this PR with those changes so we can merge it ASAP! thanks

AzazelSensei:
Rebased on the very_good_analysis 11 merge and ran dart format on the coverage test. Analyze is clean on those files.

marcossevilla:
@AzazelSensei for some reason code coverage dropped 🤔

AzazelSensei:
Build was at 99.97% on the unused absolute reportOn branch. Both relative and absolute dirs now go through path.join.

marcossevilla (approved):

Issue #804, opened 2023-09-21
fix: Include a command to generate a single coverage report for multi-package apps


3ettilina:
**Description**
After running `very_good test ... --coverage -r` there's no easy way of creating a single coverage report with the `lcov` files that were generated.

Would be great to have a new option to generate a single HTML report file whilst the tests are running and be able to open (manually or automatically) the report afterwards.

**Steps To Reproduce**
1. Run `very_good test ... --coverage -r`
2. The command generates LCOV files inside each package of the app
3. There's no in-CLI way to merge them together in a single coverage report file

**Expected Behavior**
1. Run something like `very_good test ... --coverage **--report** -r`
2. The command generates a single report with all LCOV files generated in each package
4. Can be opened by simply running `open coverage/index.html` or having another command option `--coverage-open`, which does both things (generating and opening) 🤔 

**Additional Context**
Would be great to see (if possible) the HTML report file as:
- lib
- packages/api_client
- packages/app_ui
- packages/user_repository
- ...

So we can have the whole coverage of each package in the first page, and then tapping on each of those bullet options we can navigate to the coverage of that package.

marcossevilla:
closing this since it was solved through another issue, let us know if the issue persists 👍

PR #1707, opened 2026-08-28
feat(test): support sharding tests across CI runners

peter-trost:
## Description

Adds `--shard-index` and `--total-shards` to `very_good test` and `very_good dart test`, so a test suite can be split across multiple CI runners with a `strategy.matrix`.

```yaml
strategy:
  matrix:
    shard: [1, 2, 3]
steps:
  - run: very_good test --shard-index ${{ matrix.shard }} --total-shards 3
```

Opening as a **draft** because @ryzizub is assigned to #1538 — happy to hand this over, close it, or adapt to a different design if work is already underway or a different approach is preferred.

## Why this approach

The test optimizer already discovers every test file, so sharding is a **partition of that list** rather than new machinery.

- **Sorted, then round-robin.** `Directory.listSync` order is filesystem dependent, so without sorting two runners could disagree on the partition and either skip or duplicate tests. Sorting first makes the partition deterministic across machines; dealing out round-robin keeps shards balanced by file count.
- **Non-optimized tests are sharded too.** Files tagged `skip_very_good_optimization` run as standalone files alongside the optimizer entrypoint. Leaving them unsharded would re-run all of them on every runner, partly defeating the purpose.
- **No shard-specific filenames.** The issue proposed generating `.test_optimizer_1_of_3.dart`. That turned out to be unnecessary — each runner generates its own `.test_optimizer.dart` containing only its slice, and the file is cleaned up afterwards anyway. Same result, no filename plumbing.
- **Empty shards succeed.** A shard with no test files (more shards than test files) would otherwise fail with `No tests were found`, breaking builds on oversized matrices. It now reports success, matching the existing "no test folder" behaviour.

## Interaction with coverage

`--min-coverage` is **rejected with a usage error** when sharding. Each shard only exercises a subset of the codebase, so its coverage is not representative of the whole suite and would fail the build spuriously. The error explains the alternative: collect per-shard coverage with `--coverage`, merge the lcov reports, and enforce the threshold once in a separate job.

Sharding also requires the optimizer, so it is rejected with `--no-optimization` and with `--platform` (which disables the optimizer). Both cases exit with `ExitCode.usage` and an actionable message.

This is the part I'd most like a maintainer opinion on — a documented merge-then-check workflow is the other reasonable option, and #804 may change what's possible here.

## Testing

- **Hook unit tests** (11 new): complete/disjoint partition, non-optimized sharding, determinism, balance, empty shards, and the pass-through case.
- **Command tests** (15 new): all validation paths and shard pass-through, for both `test` and `dart test`.
- **Runner tests** (2 new): shard vars reach the generator; empty shard short-circuits without invoking the test runner.
- **Manual end-to-end** against a real package (7 optimizable + 2 tagged test files): verified all tests run exactly once across 3 and 12 shards, that shards are stable across repeated runs, and that the Flutter path shards correctly. Also exhaustively verified the partition is complete, disjoint, and balanced for suite sizes 0–40 across 1–8 shards.

`dart analyze --fatal-infos --fatal-warnings .` is clean, and the touched `lib/` files are at 100% line coverage.

> [!NOTE]
> The full suite passes with `-j 1` (529 tests). At `-j 8` a handful of unrelated tests fail due to a pre-existing working-directory race between test files — I confirmed the same failures on a clean `main` checkout, so it is not introduced here.

Closes #1538

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ryzizub:
Thanks for the contribution @peter-trost ! ❤️ You can undraft it, since I actually havent started much on the sharding tests and this actually on first glance does it how i would do it. 

I will give it a test and bigger review later today/tomorrow but so far it looks good to me

peter-trost:
@ryzizub nice, glad to hear! Let me know if I can support in any way :)

peter-trost:
the flame e2e failing seems unrelated 🤔

ryzizub:
> the flame e2e failing seems unrelated 🤔

Correct, there seems to be issue unrelated to this PR

lib/src/commands/test/test.dart
```diff
@@ -452,6 +541,17 @@ This command should be run from the root of your Flutter project.''');
 
     final options = FlutterTestOptions.parse(_argResults, config: config);
 
+    final shardingError = validateSharding(
+      rawShardIndex: _argResults['shard-index'] as String?,
```
ryzizub:
```suggestion
      rawShardIndex: options.shardIndex,
      
```

lib/src/commands/test/test.dart
```diff
@@ -452,6 +541,17 @@ This command should be run from the root of your Flutter project.''');
...
+      rawTotalShards: _argResults['total-shards'] as String?,
```
ryzizub:
```suggestion
      rawTotalShards: options.totalShards,
```

peter-trost:
a505315 adds a fix to put optimized and `skip_very_good_optimization` files into one sorted list instead of two. Sharding them separately restarted at shard 1 for each list, so early shards got a file from both while later ones sat empty.

bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -33,35 +33,78 @@ Future<void> run(HookContext context) async {
...
+/// Returns the subset of [paths] that belongs to the shard [shardIndex] out of
+/// [totalShards].
+///
+/// Returns [paths] unchanged when sharding is not enabled (either value is
+/// `null`).
+///
+/// Files are dealt out round-robin (index modulo [totalShards]) over the
+/// already sorted [paths], which keeps shards balanced in file count and makes
+/// the partition stable for a given test suite.
+List<String> shardOf(
```
ryzizub: Does this needs to be public?

test/src/commands/test/test_test.dart
```diff
@@ -53,6 +53,8 @@ const expectedTestUsage = [
       '    --flavor                                                 Build a custom app flavor as defined by platform-specific build setup. Supports the use of product flavors in Android Gradle scripts, and the use of custom Xcode schemes.\n'
       '    --timeout=<seconds>                                      Maximum seconds to let tests run before killing the process. Useful when tests hang due to an unbounded pumpAndSettle() call.\n'
       '    --file-reporter=<name:path>                              Enable an additional reporter writing test results to a file. Should be in the form <name>:<path> (e.g. "json:reports/tests.json").\n'
+      '    --shard-index=<1>                                        The 1-based index of the shard to run. Must be used together with --total-shards. Requires optimization to be enabled.\n'
```
ryzizub: the value inside <> should be named not exact and in the description should be explained what is default

bricks/test_optimizer/hooks/test/pre_gen_test.dart
```diff
@@ -286,5 +287,189 @@ dependencies:
         },
       );
     });
+    group('Sharding', () {
+      /// Creates a package with [count] optimizable test files, plus any
+      /// [notOptimized] files carrying the skip optimization tag.
+      Directory createPackage(int count, {int notOptimized = 0}) {
```
ryzizub, line 312: Would be good to clean out these created files.

ryzizub (commented):
Overall LGTM, great job! Just a small findings

peter-trost:
Thanks for the review @ryzizub! Latest commit addresses the feedback.

bricks/test_optimizer/hooks/test/pre_gen_test.dart
peter-trost: These live under `tempDirectory`, which the top-level `tearDown` deletes recursively after each test, so they're already cleaned up. 

Should we still add an explicit tear down here?

ryzizub: Nono sorry, missed that. All good here

peter-trost:
@ryzizub merge conflicts fixed.

ryzizub:
Adding @marcossevilla as additional reviewer for this PR

lib/src/commands/test/test.dart
```diff
@@ -452,6 +538,17 @@ This command should be run from the root of your Flutter project.''');
 
     final options = FlutterTestOptions.parse(_argResults, config: config);
 
+    final shardingError = validateSharding(
+      rawShardIndex: options.shardIndex,
+      rawTotalShards: options.totalShards,
+      optimizePerformance: options.optimizePerformance,
```
marcossevilla: **Blocking:** this passes the raw `--optimization` flag, but the effective value at L564-570 is also switched off by `--platform`, `--update-goldens`, and an explicit test path.
So `very_good test --platform chrome --shard-index 1 --total-shards 3` passes the check at L275-277, then runs with the optimizer off and every shard runs the whole suite.
Compute the effective value once above this call and pass it to both `validateSharding` and `_flutterTest`.

lib/src/commands/dart/commands/dart_test_command.dart
```diff
@@ -374,6 +406,17 @@ This command should be run from the root of your Dart project.''');
 
     final options = DartTestOptions.parse(_argResults, config: config);
 
+    final shardingError = validateSharding(
+      rawShardIndex: options.shardIndex,
+      rawTotalShards: options.totalShards,
+      optimizePerformance: options.optimizePerformance,
```
marcossevilla: **Blocking:** same as in `test.dart`. This is the raw flag, but L431-437 also switches it off for `--platform` and for explicit test paths.
`very_good dart test --platform chrome --shard-index 1 --total-shards 4` then runs the full suite on all four shards instead of a quarter each.
Compute the effective value once and pass it to both.

lib/src/cli/test_cli_runner.dart
```diff
@@ -162,6 +168,20 @@ class TestCLIRunner {
 
         final notOptimizedTests =
             vars['notOptimizedTests'] as List<dynamic>? ?? [];
+
+        // A shard can legitimately end up empty when there are more shards
+        // than test files. That is not a failure: the shard has nothing to do,
+        // so report success instead of letting the test runner exit with
+        // "No tests were found", which would fail the CI job.
+        if (shardIndex != null &&
```
marcossevilla, line 240: **Bug:** L123-127 deletes `coverage/lcov.info`, and returning here skips the block that rewrites it, which ends by asserting the file exists (L266-267).
`very_good test --coverage --shard-index 4 --total-shards 4` on a three-file suite then succeeds with no lcov file, so the merge step has nothing to read.
Write an empty lcov file before returning.

lib/src/commands/test/test.dart
```diff
@@ -218,10 +224,71 @@ class FlutterTestOptions {
...
+  // cannot work when the optimization is turned off.
+  if (!optimizePerformance) {
+    return 'Sharding requires optimization to be enabled. '
+        'Remove --no-optimization or --platform to use sharding.';
+  }
+
+  // Each shard only exercises a fraction of the codebase, so its coverage is
+  // not representative of the whole suite. Merge the lcov files from every
+  // shard and enforce the threshold in a separate job instead.
+  if (minCoverage != null) {
```
marcossevilla: **Bug:** `minCoverage` is resolved from `very_good.yaml` (L56-60), so this rejects sharding in any repo that sets `min_coverage:`, even when the flag was never passed.
There is no `--no-min-coverage` to opt out, only `--min-coverage=""`.
Check `argResults.wasParsed('min-coverage')` instead, the same way `rawShardIndex` and `rawTotalShards` stay strings so "not provided" can be detected.

bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -33,35 +33,80 @@ Future<void> run(HookContext context) async {
...
+/// Returns the subset of [paths] that belongs to the shard [shardIndex] out of
+/// [totalShards].
+///
+/// Returns [paths] unchanged when sharding is not enabled (either value is
+/// `null`).
+///
+/// Files are dealt out round-robin (index modulo [totalShards]) over the
+/// already sorted [paths], which keeps shards balanced in file count and makes
+/// the partition stable for a given test suite.
+List<String> _shardOf(
```
marcossevilla, line 114: **Bug:** the loop below has no bounds checks. `totalShards: 0` never advances `i`, so it appends `paths[0]` until memory runs out, and `shardIndex: 0` reads `paths[-1]` and throws.
`validateSharding` blocks both from the CLI, but `brick.yaml` now prompts for these two vars, so `mason make test_optimizer` reaches this code directly.
Add a guard here, since this is the only place that can protect that caller.

bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -33,35 +33,80 @@ Future<void> run(HookContext context) async {
...
+  // optimized ones: dealing out one list keeps every shard within one file
+  // of the others, whereas dealing out the two lists separately would hand
+  // the first shards a file from each.
+  final shardPaths = _shardOf(
+    testPaths,
+    shardIndex: shardIndex,
+    totalShards: totalShards,
+  );
+  final optimizedTestPaths = shardPaths
+      .where((p) => !notOptimizedTests.contains(p))
```
marcossevilla, line 86: **Nit:** `notOptimizedTests` is a `List`, so each `contains` is a full scan, run once per path here and again on L73.
That is quadratic in the number of test files, on every run of the command.
Build a `Set` once before both `where` calls.

bricks/test_optimizer/hooks/lib/pre_gen.dart
```diff
@@ -85,9 +130,10 @@ Future<List<String>> getNotOptimizedTests(
     }
   }
 
-  /// Format to relative path
+  /// Format to relative path, normalizing separators so the paths compare
+  /// equal to the ones built in [run] on Windows too.
   final relativePaths = testWithVeryGoodTest
-      .map((e) => path.relative(e, from: testDir))
+      .map((e) => path.relative(e, from: testDir).replaceAll(r'\', '/'))
```
marcossevilla, line 152: **Nit:** this normalizes to forward slashes, but the consumer at `test_cli_runner.dart` L209 still calls `p.join('test', e.toString())`, so on Windows the argument becomes `test\sub/foo_test.dart`.
Windows accepts mixed separators, so it most likely works.
Use `p.joinAll(p.posix.split(e))` there if you want the emitted path to stay native.

lib/src/commands/dart/commands/dart_test_command.dart
```diff
@@ -7,6 +7,7 @@ import 'package:mason/mason.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:very_good_cli/src/cli/cli.dart';
+import 'package:very_good_cli/src/commands/test/test.dart';
```
marcossevilla: **Nit:** this import exists only to reach `validateSharding`, which sits next to `FlutterTestOptions` in the Flutter command.
`CLAUDE.md` rules out sibling dependencies in the same layer, and the function has no Flutter specifics.
Move it to `lib/src/cli/`, next to `TestCLIRunner`.

lib/src/commands/test/test.dart
peter-trost: Fixed. The effective value is now computed once and shared.

lib/src/commands/dart/commands/dart_test_command.dart
peter-trost: Fixed. The effective value is now computed once and shared.

lib/src/cli/test_cli_runner.dart
peter-trost: Fixed. An empty lcov.info is written.

lib/src/commands/test/test.dart
peter-trost: Agreed on the problem. One tweak: wasParsed alone would let a config threshold through to per-shard enforcement and fail spuriously. So we reject only an explicit flag, and drop the config value while sharding.

bricks/test_optimizer/hooks/lib/pre_gen.dart
peter-trost: Done

bricks/test_optimizer/hooks/lib/pre_gen.dart
peter-trost: Done

bricks/test_optimizer/hooks/lib/pre_gen.dart
peter-trost: Done

lib/src/commands/dart/commands/dart_test_command.dart
peter-trost: Done

peter-trost:
Thanks for the thorough review @marcossevilla. Addressed all comments. 🙂

peter-trost:
@marcossevilla resolved conflicts. Ready for your re-review :)

marcossevilla:
@peter-trost holding a bit on reviewing this since we had large changes for how optimization is handled inside the CLI, we addressed the issue on preserving test annotations and that should be the last.

sorry to cause so many changes on this PR, please lmk when the conflicts are solved and I can give a fresh review 😄

peter-trost:
@marcossevilla no worries! :D The changes are integrated, please have another look :)

peter-trost:
@marcossevilla attempted a fix for the failing workflow on windows. Should pass now. 🤞

peter-trost:
@marcossevilla friendly ping :)

marcossevilla:
hey @peter-trost, your PR lgtm but we're going through local testing for it atm. will ping you when that's wrapped up, sorry for the delays 🙏

lib/src/cli/test_cli_runner.dart
```diff
@@ -145,6 +204,17 @@ class TestCLIRunner {
           logger: logger,
         );
 
+        if (optimization.isEmptyShard) {
```
marcossevilla, line 207: this handles a shard that has no files but not a shard whose files all get filtered out by --exclude-tags. 

in that case flutter test exits 70, so the shard fails the build even though nothing should've run.
 
after trying on an app with golden tests, with -x golden across 3 shards, any shard that ended up golden-only failed. could we treat "zero tests ran" as success here too?

peter-trost: Fixed in https://github.com/VeryGoodOpenSource/very_good_cli/pull/1707/changes/4f7a1a355759fd40b6cb89dcca245c50411c7e0f

That's a nice catch, thanks for flagging!

PR #1725, opened 2026-09-09
fix(mcp): transient test failures

### LINKED ISSUE #1722: fix(mcp): test failures are classified as `transient`, telling agents to retry a deterministic failure
## Description

`ToolFailureType.fromExitCode` (added in #1687, shipped in 1.5.0) maps `ExitCode.unavailable` (69) to `ToolFailureType.transient`. But `very_good test` exits 69 for the two most common outcomes of the `test` tool — **tests ran and failed**, and **coverage came in under `--min-coverage`**.

Both therefore return `"failureType": "transient"` together with:

```json
"alternativeApproaches": [
  "Retry the command; the failure may resolve on its own.",
  "Check network and remote service availability, then retry.",
  "If retries keep failing with the same error, switch approach."
]
```

Neither failure is transient. Both are perfectly deterministic, and the advice tells an agent loop to re-run the suite instead of reading `partialResults` and fixing the code — the opposite of what #1615 set out to achieve ("so the agent loop doesn't take longer or fails unexpectedly due to missing error context").

By #1615's own category definitions this should be `business`:

> - transient (timeout, rate limit — may succeed on retry)
> - business (rule violation — escalate or find alternative)

A failing assertion or an unmet coverage threshold is a rule violation, not a timeout. `business` is also `fromExitCode`'s documented default for outcomes that cannot be attributed to the other three.

## Steps To Reproduce

1. Create a minimal Dart package with one failing test:

   ```dart
   // test/probe_test.dart
   import 'package:test/test.dart';
   void main() {
     test('fails on purpose', () => expect(1 + 2, 99));
   }
   ```

2. Call the MCP `test` tool against it: `{"directory": "<pkg>", "dart": true}`
3. Observe the returned payload:

   ```json
   {
     "status": "partial_failure",
     "failureType": "transient",
     "reason": "failed with exit code 69.",
     "partialResults": "... Expected: <99>\n  Actual: <3> ...",
     "alternativeApproaches": ["Retry the command; the failure may resolve on its own.", "..."]
   }
   ```

The same happens for a coverage shortfall — `{"directory": "<pkg>", "dart": true, "show_uncovered": true, "min_coverage": "100"}` against a partially covered package also yields `"failureType": "transient"`.

For contrast, a genuine usage error is classified correctly: passing `recursive: true` together with `paths: [...]` exits 64 and yields `"failureType": "validation"`.

## Expected Behavior

A test run that completed and reported failures — or a coverage check that ran and came in under threshold — should be classified `business` (or a dedicated bucket), with `alternativeApproaches` pointing the agent at `partialResults` rather than suggesting a retry.

## Additional Context

Verified against very_good_cli 1.5.0 (installed via `dart pub global activate very_good_cli`) on macOS, Dart 3.13.2, driving the MCP server through the `vgv-ai-flutter-plugin` `very-good-cli` server entry.

The root difficulty is that exit 69 is overloaded — it covers both "the suite ran and failed" and "the suite could not run" (see also #392, where a warning during a test produced 69). Distinguishing those inside the test command, rather than inferring intent from the exit code alone, would let `fromExitCode` keep its current sysexits mapping for every other tool while `test` reports the bucket it actually means.

Related: #1615 (the originating request), #1687 (the implementation), #1611/#1612 (the earlier output-capture fix that made `partialResults` useful in the first place).

marcossevilla:
## Status

**READY**

## Description

Closes #1722 

Fixes the status codes for transient failures in MCP.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


marcossevilla:
@vgvbot fix CI failures

marcossevilla:
hey @jmadren, adding you as reviewer since you opened the issue initially, lmk if it fixes your problem 👍

erickzanardo (approved):

lib/src/commands/dart/commands/dart_test_command.dart
```diff
@@ -379,62 +379,63 @@ This command should be run from the root of your Dart project.''');
     final config = VeryGoodConfig.load(Directory(targetPath), logger: _logger);
     if (config == null) return ExitCode.config.code;
 
-    final isDartInstalled = await _dartInstalled(logger: _logger);
+    if (!await _dartInstalled(logger: _logger)) {
+      return ExitCode.success.code;
```
ryzizub, line 383: Should this be success?

lib/src/cli/test_cli_runner.dart
```diff
@@ -301,12 +301,11 @@ class TestCLIRunner {
...
       : overrideAnsiOutput(enableAnsiOutput, body);
 
-  /// Handles the [MinCoverageNotMet] exception by logging the error message.
-  ///
-  /// If [e] contains uncovered lines, they are logged after the error message.
-  static void handleMinCoverageNotMet({
+  /// Logs [error], along with its uncovered lines when it carries any, and
+  /// returns the exit code an unmet coverage threshold reports.
+  static int handleMinCoverageNotMet(
+    MinCoverageNotMet error, {
```
ryzizub, line 307: why we moved from positional arguments here?

lib/src/commands/test/test.dart
```diff
@@ -457,72 +457,73 @@ This command should be run from the root of your Flutter project.''');
     final config = VeryGoodConfig.load(Directory(targetPath), logger: _logger);
     if (config == null) return ExitCode.config.code;
 
-    final isFlutterInstalled = await _flutterInstalled(logger: _logger);
+    if (!await _flutterInstalled(logger: _logger)) {
+      return ExitCode.success.code;
```
ryzizub, line 461: same https://github.com/VeryGoodOpenSource/very_good_cli/pull/1725/changes#r3980503066

lib/src/commands/dart/commands/dart_test_command.dart
marcossevilla: this was the behavior already, I just moved it as a guard to reduce nesting, but we could return another exitcode like unavailable if needed, it's just not the scope of this PR

lib/src/cli/test_cli_runner.dart
marcossevilla: the only positional is the error, the other ones are named parameters because one is optional and the other is not the actual error. changed it for readability but we can revert

lib/src/commands/dart/commands/dart_test_command.dart
ryzizub: I think unavailable would be more proper, we can change it as a follow up

PR #1726, opened 2026-09-09
feat(test): exclude test files from the optimized bundle

### LINKED ISSUE #1713: feat(test): allow excluding files from optimized bundles in very_good.yaml
## Problem

`very_good test --optimization` always collects every `*_test.dart` file under `test/` unless the file contains the special `skip_very_good_optimization` tag. Projects cannot exclude a file or directory from the generated optimized bundle through `very_good.yaml`.

This becomes especially surprising with file-level `@Tags`: the generated `.test_optimizer.dart` imports each suite and invokes its `main()` inside an untagged `group(...)`. A command such as `very_good test --optimization --exclude-tags integration` therefore cannot keep a file-level `@Tags(['integration'])` suite out of the optimized bundle.

Very Good CLI 1.4.0 still generates the untagged wrapper in `bricks/test_optimizer/__brick__/test/.test_optimizer.dart`, while `bricks/test_optimizer/hooks/lib/pre_gen.dart` filters files only by scanning for `skip_very_good_optimization`.

## Proposal

Add an optional glob list to the `test` section of `very_good.yaml` for files that must run outside the optimized bundle. For example:

```yaml
test:
  optimization: true
  optimization_exclude:
    - test/integration/**
    - test/**/serial_*_test.dart
```

Matching files would be omitted from `.test_optimizer.dart` and passed to the underlying test runner separately, in the same way files marked `skip_very_good_optimization` are handled today. The exact key name is flexible; the important part is a declarative path/glob exclusion in `very_good.yaml`.

## Why this is separate from tag propagation

File-level tag support was declined in #576 and #522 because fully honoring package:test tags would require boolean-selector evaluation, include/exclude conflict semantics, source parsing, and `dart_test.yaml` defaults.

This proposal does not parse or interpret Dart tags. Matching paths before generating the bundle is deterministic, avoids tag-selector semantics, and provides a project-wide alternative to placing a source marker in every excluded file.

This follows the direction discussed in #1609: a maintainer recommended a YAML property for optimization exclusions and directed the requirement to #360. PR #1636 subsequently introduced `very_good.yaml`, but the released test configuration does not include an optimization path-exclusion option.

## Acceptance criteria

- `very_good.yaml` accepts a documented list of file or directory patterns under `test`.
- Matching test files are not imported by the optimized bundle.
- Matching files still run outside the bundle rather than being skipped entirely.
- Existing `skip_very_good_optimization` behavior remains backward-compatible.
- Tests cover individual file patterns, directory patterns, and non-matching patterns.

marcossevilla:
Adds `optimization.exclude` to the `test` and `dart.test` sections of
`very_good.yaml`, plus a repeatable `--exclude-optimization` option on
`very_good test` and `very_good dart test`. Matching test files are left out
of the generated `.test_optimizer.dart` and handed to the test runner as
their own targets, which also restores file-level `@Tags` handling for them.

`optimization` now accepts a map with `enabled` and `exclude`. The previous
boolean form is shorthand for `enabled`, so `optimization: false` keeps
working.

The optimized entrypoint is no longer passed to the runner once every test
has been excluded from it, unless there is nothing else left to run.

Closes #1713


erickzanardo (dismissed):

ryzizub (dismissed):
LGTM

marcossevilla:
@vgvbot fix CI failures

lib/src/cli/cli.dart
```diff
@@ -10,9 +10,14 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:pubspec_parse/pubspec_parse.dart';
 import 'package:universal_io/io.dart';
-import 'package:very_good_cli/src/cli/templates/templates.dart';
+import 'package:very_good_cli/src/test_optimizer/test_optimizer.dart';
 import 'package:very_good_test_runner/very_good_test_runner.dart';
 
+// The optimizer is part of this library's surface: `Flutter.test`, `Dart.test`
+// and `TestCLIRunner.test` all take one, so every importer of `cli.dart` needs
+// its types without reaching for a second import.
```
marcossevilla:
```suggestion
```

ryzizub (approved):
LGTM

